PR #4, opened 2018-08-19
Implement co-log interface

ChShersh:
This PR implements general idea of interface for `co-log` and some useful combinators.

### DIFF
--- a/co-log/example/Main.hs
+++ b/co-log/example/Main.hs
@@ -15,5 +15,10 @@ app = do
     logMsg "Starting application..."
     withLog (cmap ("app:" ++)) example
 
+foo :: (WithLog env String m, WithLog env Int m) => m ()
+foo = do
+    logMsg "String message..."
+    logMsg @Int 42
+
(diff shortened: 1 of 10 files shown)


co-log/example/Main.hs
```diff
@@ -0,0 +1,24 @@
...
+    logMsg "First message..."
+    logMsg "Second message..."
+
+app :: WithLog env String m => m ()
+app = do
+    logMsg "Starting application..."
+    withLog (cmap ("app:" ++)) example
+
+foo :: (WithLog env String m, WithLog env Int m) => m ()
+foo = do
```
vrom911: Is this used anywhere?

ChShersh: It's not used currently. I've added this example just to check whether it builds or not. To actually launch this example I need to create some data type `Env` and implement the following instances:

```haskell
instance HasLog Env String m
instance HasLog Env Int m
```

This is more complicated so I decided to have only example and I hope to run it in future...

PR #12, opened 2018-09-11
[#6] Add severity and LogMessage with severity

### LINKED ISSUE #6: Add `Severity` data type
It's useful for message filtering by severity.

vrom911:
Resolves #6

### DIFF
--- a/co-log-core/src/Colog/Core/Severity.hs
+++ b/co-log-core/src/Colog/Core/Severity.hs
@@ -1,0 +1,24 @@
+module Colog.Core.Severity
+       ( Severity (..)
+       , filterSeverity
+       ) where
+
+import Colog.Core.Action (LogAction (..), cfilter)
+
+-- | Severity for the log messages.
+data Severity
+    -- | Information useful for debug purposes
+    = Debug
+    -- | Normal operational information
+    | Info
+    -- | Important (more than average) but not critical information
+    | Notice
+    -- | General warnings
+    | Warning
+    -- | General errors/severe errors
+    | Error
+    deriving (Show, Read, Eq, Ord, Enum, Bounded)
+
+-- Filters by given 'Severity'.
+filterSeverity :: Applicative m => Severity -> (a -> Severity) -> LogAction m a -> LogAction m a
+filterSeverity s fs = cfilter (\a -> fs a >= s)
--- a/co-log-core/src/Colog/Core/Severity.hs
+++ b/co-log-core/src/Colog/Core/Severity.hs
@@ -1,24 +1,57 @@
+{-# LANGUAGE PatternSynonyms #-}
+
+{- | This module introduces 'Severity' data type for expressing how severe the
+message is. Also, it contains useful functions for work with 'Severity'.
+
+-}
+
 module Colog.Core.Severity
        ( Severity (..)
-       , filterSeverity
+         -- Patterns
+       , pattern D
+       , pattern I
+       , pattern W
+       , pattern E
+       , filterBySeverity
        ) where
+
+import Data.Ix (Ix)
 
 import Colog.Core.Action (LogAction (..), cfilter)
 
 -- | Severity for the log messages.
 data Severity
-    -- | Information useful for debug purposes
-    = Debug
-    -- | Normal operational information
-    | Info
-    -- | Important (more than average) but not critical information
-    | Notice
-    -- | General warnings
-    | Warning
-    -- | General errors/severe errors
-    | Error
-    deriving (Show, Read, Eq, Ord, Enum, Bounded)
+    {- | Information useful for debug purposes.
 
--- Filters by given 'Severity'.
-filterSeverity :: Applicative m => Severity -> (a -> Severity) -> LogAction m a -> LogAction m a
-filterSeverity s fs = cfilter (\a -> fs a >= s)
+    E.g. output of the function that is important for the internal development,
+    not for users. Like, the result of SQL query.
+    -}
+    = Debug
+    {- | Normal operational information.
+
+    E.g. describing general steps: starting application, finished downloading.
+    -}
+    | Info
+    {- | General warnings, non-critical failures.
+
+    E.g. couldn't download icon from some service to display.
+    -}
+    | Warning
+    {- | General errors/severe errors.
+
+    E.g. exceptional situations: couldn't syncronize accounts.
+    -}
+    | Error
+    deriving (Show, Read, Eq, Ord, Enum, Bounded, Ix)
+
+pattern D, I, W, E :: Severity
+pattern D <- Debug   where D = Debug
+pattern I <- Info    where I = Info
+pattern W <- Warning where W = Warning
+pattern E <- Error   where E = Error
+{-# COMPLETE D, I, W, E #-}
+
+
+-- | Filters messages by given 'Severity'.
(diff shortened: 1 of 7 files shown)


co-log-core/src/Colog/Core/Severity.hs
```diff
@@ -0,0 +1,24 @@
...
+    -- | Information useful for debug purposes
+    = Debug
+    -- | Normal operational information
+    | Info
+    -- | Important (more than average) but not critical information
+    | Notice
+    -- | General warnings
+    | Warning
+    -- | General errors/severe errors
+    | Error
```
ChShersh: What do you think about possible changes to this data type? I want to say that I'm quite happy with current data type. But I would like to discuss with you possible ideas I have in mind (though, not of them are good).

1. Other name than `Severity`? Like `Level` or `LogLevel` or `Verbosity`? I like `Severity` but maybe some other name might fit better.
2. Add more severities? Like `Critical`? Or remove severity (like `Notice`: I personally find hard to distinguish when I should print something with `Info` or `Notice`?
3. Following after last question: Maybe we should add to documentation examples of when users need to use levels? For example, 

    * `Debug` for then you want to show output of function that is important for internal development, not for users. Like, result of SQL query.
    * `Info` for describing general steps: starting application, finished downloading.
    * `Notice` for important message: don't forget to update settings, the size of your local storage becomes bigger than average.
    * `Warning` for non-critical failures: couldn't download icon from some service to display
    * `Error` for exceptional situations: couldn't syncronize accounts.

    Maybe you have better ideas for examples 🙂 
4. What do you think about introducing bidirectional single-letter pattern-synonyms for these constructors? There're couple reasons for having additional pattern-synonyms:
    1. Using fancy feature.
    2. Shorter alternative names: we give users multiple ways to use our data type.
    3. Good alignment. For example instead of:

```haskell
log Debug "Foo"
log Warning "Bar"
```

users can write:

```haskell
log D "Foo"
log W "Bar"
```

If you think, that this is a good idea, you can take inspiration from pattern synonym in `tomland` library:

* https://github.com/kowainik/tomland/blob/db96cd45b3b9f04db7ce96c630b160764775c148/src/Toml/PrefixTree.hs#L80-L85

co-log-core/src/Colog/Core/Severity.hs
```diff
@@ -0,0 +1,24 @@
...
+    deriving (Show, Read, Eq, Ord, Enum, Bounded)
```
ChShersh: I really like that we have all these derivings! This will make `Severity` data type easier to use with `relude` and functions like `universe` and `inverseMap` 👍 

Let's also add `Ix` to `deriving`. Just in case:

* https://hackage.haskell.org/package/base-4.11.1.0/docs/Data-Ix.html

co-log-core/src/Colog/Core/Severity.hs
```diff
@@ -0,0 +1,24 @@
...
+
+-- Filters by given 'Severity'.
+filterSeverity :: Applicative m => Severity -> (a -> Severity) -> LogAction m a -> LogAction m a
```
ChShersh: Maybe more suitable name for this function is `filterBySeverity`? 🤔

co-log-core/src/Colog/Core/Severity.hs
```diff
@@ -0,0 +1,24 @@
+module Colog.Core.Severity
```
ChShersh: Let's add some short top-level documentation to this module as well.

co-log/example/Main.hs
```diff
@@ -1,26 +1,30 @@
...
+import Colog (LogMessage (..), WithLog, cmap, fmtLogMessage, logDebug, logInfo, logMsg,
+              logTextStderr, logTextStdout, logWarning, usingLoggerT, withLog, withLogTextFile)
 
-example :: WithLog env String m => m ()
+example :: WithLog env LogMessage m => m ()
 example = do
-    logMsg "First message..."
-    logMsg "Second message..."
+    logDebug "First message..."
+    logInfo "Second message..."
```
ChShersh: I would like to ask you what style of logging do you prefer?

1. `logDebug "Message"`
2. `log Debug "Message"`.

The first style is one character shorter. But you need to import a lot of different functions if you're using logging severity of different levels: 
```haskell
import Colog (logDebug, logWarn, logInfo, logError, logNotice)
```
The second style is one character longer, but imports are shorter:
```haskell
import Colog (Severity (..), log)
```

Maybe there're other cons and pros of these styles? 🤔 But even if we stick to single style, it might be a good idea to provide to users functions of different types.

co-log/example/Main.hs
```diff
@@ -1,26 +1,30 @@
...
-app :: WithLog env String m => m ()
+app :: WithLog env LogMessage m => m ()
 app = do
-    logMsg "Starting application..."
-    withLog (cmap ("app:" ++)) example
+    logWarning "Starting application..."
+    withLog (cmap addApp) example
+  where
+    addApp :: LogMessage -> LogMessage
+    addApp (LogMessage msg sev) = LogMessage ("app: " <> msg) sev
```
ChShersh: That's the correct way to handle `cmap` for data types. But with lens this can be written shorter without need to introduce new function each time, like:

```haskell
withLog (cmap $ msg %~ ("app:" <>)) example
```

(though, some people might say that the above code is not really readable)

I think that we will add some lens library to `colog` at some point in future. I haven't decided yet, which one: `lens` or `microlens`. But we will figure out it later.

co-log/src/Colog/Message.hs
```diff
@@ -0,0 +1,45 @@
...
+       , logWarning
+       , logError
+
+       , fmtLogMessage
+       ) where
+
+import Colog.Core.Severity (Severity (..))
+import Colog.Monad (WithLog, logMsg)
+
+data LogMessage = LogMessage Text Severity
```
ChShersh: What do you think about renaming this data type from `LogMessage` to just `Message`? I think this looks nice in constraints:

```haskell
WithLog env Message m => ...
```

And we can have nice record data type:

```haskell
data Message = Message
    { messageSeverity :: {-# UNPACK #-} !Severity
    , messageText     :: {-# UNPACK #-} !Text
    } deriving ... something useful ... ?
```

(those `{-# UNPACK #-}` and `!` annotations are for performance, because we want performant logging library)

co-log/src/Colog/Message.hs
```diff
@@ -0,0 +1,45 @@
...
+-- | Logs the message with 'Warning' severity.
+logWarning :: WithLog env LogMessage m => Text -> m ()
+logWarning = log Warning
+
+-- | Logs the message with 'Error' severity.
+logError :: WithLog env LogMessage m => Text -> m ()
+logError = log Error
+
+fmtLogMessage :: LogMessage -> Text
+fmtLogMessage (LogMessage msg sev) = "[" <> show sev <> "] " <> msg
```
ChShersh: Later we will add fancy coloring and formatting options here, so our log messages will become really nice ✨

co-log/example/Main.hs
vrom911: Yes, I'm not sure which I like more, I just get used to the style above, but I agree that we should give users the opportunity to choose their favourite way, it won't harm.

PR #16, opened 2018-09-12
[#9] Add Contravariant, Divisible, Decidable instances

### LINKED ISSUE #9: Add `Contravariant`, `Divisible` and `Decidable` instances for `LogAction`
I propose to add functions with types to `Colog.Core.Action` module in `co-log-core` package. And then create `Colog.Contra` module inside `co-log` package with orphan instances. The reasons for this is to not depend on `contravariant` library in `core` package.

vrom911:
Resolves #9

### DIFF
--- a/co-log-core/src/Colog/Core/Action.hs
+++ b/co-log-core/src/Colog/Core/Action.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE ViewPatterns #-}
+
 {- | Implements core data types and combinators for logging actions.
 -}
 
@@ -11,7 +13,16 @@ module Colog.Core.Action
          -- * Contravariant combinators
        , cfilter
        , cmap
+       , (>$)
        , cbind
+
+         -- * Divisible combinators
+       , divide
+       , conquer
+
+         -- * decidable combinators
+       , lose
+       , choose
 
          -- * Comonadic combinators
        , extract
@@ -25,6 +36,7 @@ import Data.Foldable (for_)
 import Data.List.NonEmpty (NonEmpty (..))
 import Data.Monoid (Monoid (..))
 import Data.Semigroup (Semigroup (..), stimesMonoid)
+import Data.Void (Void, absurd)
 
 ----------------------------------------------------------------------------
 -- Core data type with instances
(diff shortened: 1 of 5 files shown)


co-log-core/src/Colog/Core/Action.hs
```diff
@@ -154,6 +165,14 @@ cmap :: (a -> b) -> LogAction m b -> LogAction m a
 cmap f (LogAction action) = LogAction (action . f)
 {-# INLINE cmap #-}
 
+{- | This combinator is @>$@ from contravariant functor. Replaces all locations
+in the output with the same value. The default definition is
+@contramap . const@, so this is a more efficient version.
+-}
+infixl 4 >$
+(>$) :: b -> LogAction m b -> LogAction m a
+(>$) f (LogAction action) = LogAction (const (action f))
```
ChShersh: I guess you can rename `f` to `b`. And I think that in order to make this function more efficient than `contramap . const`, it's better to write lambda instead of using `const` function. Though, the performance difference might not be even noticeable. Though, still, it's good to do it explicitly.

co-log-core/src/Colog/Core/Action.hs
```diff
@@ -239,3 +258,20 @@ infixr 1 <<=
 (<<=) :: Semigroup msg => (LogAction m msg -> m ()) -> LogAction m msg -> LogAction m msg
```
ChShersh: Could you please move `Comonadic` combinators at the end of the module? To make this module body more consistent with reexport list.

co-log-core/src/Colog/Core/Action.hs
```diff
@@ -11,8 +13,17 @@ module Colog.Core.Action
...
        , cfilter
        , cmap
+       , (>$)
        , cbind
 
+         -- * Divisible combinators
+       , divide
+       , conquer
+
+         -- * decidable combinators
```
ChShersh: Capital letter can be here

co-log/src/Colog/Contra.hs
```diff
@@ -0,0 +1,27 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+{- | This module contains 'LogAction' instances of @contravariant@ classes.
+-}
+
+module Colog.Contra
```
ChShersh: This module can be imported from `Colog` module in `co-log` package

PR #35, opened 2018-09-20
Allow usage of MessageField with OverloadedLabels

ChShersh:


co-log/src/Colog/Message.hs
```diff
@@ -127,27 +131,31 @@ showSourceLoc cs = square showCallStack
 types. The type family is open so you can add new instances.
 -}
 type family FieldType (fieldName :: Symbol) :: Type
-type instance FieldType "thread-id" = ThreadId
-type instance FieldType "utc-time"  = UTCTime
+type instance FieldType "threadId" = ThreadId
+type instance FieldType "utcTime"  = UTCTime
```
ChShersh: Maybe I should name strings in the same way as types to make implementation better compatible with the ecosystem? 🤔

co-log/src/Colog/Message.hs
```diff
@@ -166,14 +174,14 @@ type FieldMap (m :: Type -> Type) = TypeRepMap (MessageField m)
...
+"threadId" -> myThreadId
+"utcTime"  -> getCurrentTime
 @
 -}
 defaultMessageMap :: MonadIO m => FieldMap m
 defaultMessageMap = fromList
-    [ TM.WrapTypeable $ MessageField @"thread-id" (liftIO myThreadId)
-    , TM.WrapTypeable $ MessageField @"utc-time" (liftIO getCurrentTime)
+    [ #threadId (liftIO myThreadId)
+    , #utcTime (liftIO getCurrentTime)
```
ChShersh: This is the main motivating change in this PR

vrom911: Wow! That's magic!

co-log/src/Colog/Message.hs
vrom911: I agree with the same name as the type 👍

ChShersh: I just realised that it's not possible to give the same name as type because labels can't start with capital letter 🙂 So let's keep lowerCamelCase

PR #43, opened 2018-09-30
Rename cbind to cmapM - #38

### LINKED ISSUE #38: Rename `cbind` to `cmapM`

Cmdv:
fixes #38

### DIFF
--- a/co-log-core/CHANGELOG.md
+++ b/co-log-core/CHANGELOG.md
@@ -4,8 +4,11 @@ Change log
 `co-log-core` uses [PVP Versioning][1].
 The change log is available [on GitHub][2].
 
-0.0.0.1
-=======
+0.1.0
+=====
+
+* [#38](https://github.com/kowainik/co-log/issues/38):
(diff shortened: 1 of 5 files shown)


Cmdv:
@vrom911 ah yes not a problem I'll add those now 👍

co-log-core/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ Change log
 `co-log-core` uses [PVP Versioning][1].
 The change log is available [on GitHub][2].
 
+0.0.0.2
```
vrom911: This should also be `0.1.0` to match what we have in the `.cabal` (the real version 🙂)
Also, as `0.0.0.1` is not released you can put all the changes from there (which are in the below section) in `0.1.0` as well.

Cmdv:
semver always catches me out!! but think it should be ok now 😄

Cmdv:
whoop one step at a time 🎉

PR #45, opened 2018-10-01
Introduce basic approach to concurrent log writing.

qnikst:
Introduce two approaches for concurrent log writing
and basic utilities for working with concurrent logs.
Two approaches are needed in order to allow user to
chose between generality and speed, though approaches
were not benchmarked to check the hypothesys.

### DIFF
--- a/co-log/src/Colog/Concurrent.hs
+++ b/co-log/src/Colog/Concurrent.hs
@@ -1,0 +1,318 @@
+-- |
+-- For the speed reasons you may want to dump logs asynchronously.
+-- This is especially useful when application threads are CPU
+-- bound while logs emitting is I/O bound. This approach
+-- allows to mitigate bottlenecks from the I/O.
+--
+-- When writing an application user should be aware of the tradeoffs
+-- that concurrent log system can provide, in this module we explain
+-- potential tradeoffs and describe if certain building blocks are
+-- affected or not.
+--
+--   1. Unbounded memory usage - if there is no backpressure mechanism
+--   the user threads, they may generate more logs that can be
+--   written in the same amount of time. In those cases messages will
+--   be accumulated in memory. That will lead to extended GC times and
+--   application may be killed by the operating systems mechanisms.
+--
+--   2. Persistence requirement - sometimes application may want to
+--   ensure that logs were written before it can continue. This is not
+--   a case with concurrent log systems in general, and some logs may
+--   be lost when application exits before dumping all logs.
+--
+--   3. Non-precise logging - sometimes it may happen that there can be
+--   logs reordering (in case if thread was moved to another capability).
+--
+-- In case if your application is a subject of those problems you may
+-- consider not using concurrent logging system in other cases concurrent
+-- logger may be a good default for you.
+--
+module Colog.Concurrent
+       ( -- $general
+         -- * Simple API.
+         -- $simple-api
+         withBackgroundLogger
+       , defCapacity
+         -- * Extended API
+         -- $extended-api
+         -- ** Background worker
+         -- $background-worker
+       , BackgroundWorker
+       , backgroundWorkerWrite
+       , killBackgroundLogger
+         -- ** Background logger
+       , forkBackgroundLogger
+       , convertToLogAction
+         -- ** Worker thread
+         -- $worker-thread 
+       , mkBackgroundThread
+       , runInBackgroundThread
+         -- *** Usage example
+         -- $worker-thread-usage
+       ) where
+
+import Control.Concurrent
+import Control.Concurrent.STM (check)
+import Control.Concurrent.STM.TBQueue
+import Control.Exception (bracket, finally)
+
+import Colog.Core.Action (LogAction (..))
+import Colog.Concurrent.Internal
+
+-- $general
+--
+-- Concurrent logger consists of the basic parts (see schema below).
+--
+--   1. Logger in application thread. This logger is evaluated in the
+--   application thread and has an access to all the context available
+--   in that thread and monad, this logger can work in any `m`.
+--   
+--   2. Communication channel with backpressure support. In addition to
+--   the channel we have a converter that puts the user message to the
+--   communication channel. This converter works in the user thread.
+--   Such a logger usually works in `IO` but it's possible to make it
+--   work in `STM` as well. At this point library provides only `IO`
+--   version, but it can be lifted to any `MonadIO` by the user. 
+--
+--   3. Logger thread. This is the thread that performs actual write to
+--   the sinks. Loggers there do not have access to the users thread
+--   state, unless that state was passed in the message.
+-- 
+-- 
+-- @
+--
+--  +-------------------------+                  +--------------------------------+
+--  |                         |                  | Logger        |   Sink-1       |
+--  |   Application Thread    |                  | Thread    +--->                |
+--  |   -----------------     |  +-----------+   |           |   +----------------+
+--  |                         |  |           |   +---------+ |   +----------------+
+--  |           +-------------+  |  channel  |   | Shared  +----->   Sink-2       |
+--  |           | application||  |          +----> logger  | |   |                |
+--  |           | logger    +----->          |   +---------+ |   +----------------+
+--  |           +-------------+  |           |   |           |   +----------------+
+--  |                         |  +-----------+   |           +--->   Sink3        |
+--  |                         |                  |               |                |
+--  |                         |                  |               +----------------+
+--  |                         |                  |                                |
+--  +-------------------------+                  +--------------------------------+
+-- @
+--
+-- So usually user should write the logging system in the way that all 'LogAction'
+-- that populate and filter information should live in the application logger.
+-- All loggers that do serialization and formatting should live in shared logger.
+--
+--
+-- If more concurrency is needed it's possible to build multilayer systems:
+--
+--
+-- @
+--   +-------------+                         +-------+
+--   | application |---+                 +---| sink-1|
+--   +-------------+   |   +---------+   |   +-------+
+--                     +---| logger  |---+
+--                         +---------+   |   +-------+
+--                                       +---| sink-2|
+--                                           +-------+
+-- @
+-- 
+-- In this approach application will be concurrently write logs to the logger, then
+-- logger will be concurrently writing to all sinks.
+
+
+-- $simple-api
+--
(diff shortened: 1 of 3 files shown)


qnikst:
I'm not sure about the best approach here, so I came with two, both seem to make sense and can be further extended. It's really hard to get into the trap of overgeneralization here :/

co-log/src/Colog/Concurrent.hs
```diff
@@ -0,0 +1,129 @@
+-- |
+-- Sometimes an application may need to dump logs asynchonously.
```
ChShersh: The documentation for this module is super cool!

co-log/src/Colog/Concurrent.hs
```diff
@@ -0,0 +1,129 @@
...
+--         $ \thread -> 
+--            let la1 = runInBackgroundThread worker logger1
+--                la2 = runInBackgroundThread worker logger1
+--            unLogger la1 msg1
+--            unLogger la2 msg2
+--            ..
+-- @
+--
+-- This code will run a background thread and process different
+-- loggers in that thread.
```
ChShersh: Maybe add some note like _This can be useful if logging a single message requires some computation_? Or general guideline when people might want to use concurrent logging. I personally aware only of the cases when people might not want to use concurrent logging due to some subtle concurrency issues. But it's still very useful.

co-log/src/Colog/Concurrent.hs
```diff
@@ -0,0 +1,129 @@
...
+import Control.Concurrent
+import Control.Concurrent.STM.TBQueue
+
+import Colog.Core.Action (LogAction (..))
+import Colog.Concurrent.Internal
+
+-- $worker-thread
+-- Worker thread approach allows to create a single worker
+-- thread that will perform all logging. This approach allows
+-- user to use different loggers and change them on the flight.
```
ChShersh: > and change them on the flight

That's a really cool feature — to be able to replace logging function with different one. Am I right that it's possible to do something like this: implement simple program that prints messages every second in some format `A` and listens input and after user gives some input to this program it instantly starts to print messages in format `B`?

Also, it's not that clear from existing documentation and code, how to replace current logging function with different one on the flight...

co-log/src/Colog/Concurrent.hs
```diff
@@ -0,0 +1,129 @@
...
+--
+-- With bracket resource control style code may look as
+-- following:
+--
+-- @
+-- bracket ('mkBackgroundThread' 4096)
+--         ('killBackgroundThread')
+--         $ \thread ->
```
ChShersh: Looks like `thread` variable in the documentation should be named as `worker` according to the code below

co-log/src/Colog/Concurrent.hs
```diff
@@ -0,0 +1,129 @@
...
+-- You may extend the logger and entire logger will be run
+-- on the background thread. It is not a case with
+-- 'forkBackgroundLogger'.
+
+-- | Create a background worker with a given capacity.
+-- If capacity is reached, then the thread that tries to
+-- write logs will be blocked.
+mkBackgroundThread :: Int -> IO (BackgroundWorker (IO ()))
+mkBackgroundThread capacity = do
+  queue <- newTBQueueIO (fromIntegral capacity)
```
ChShersh: I see that `newTBQueueIO` function takes `Natural` in `stm-2.5.0.0`. Maybe it's a good idea to have `stm ^>= 2.5.0.0` in `build-depends` in that case and pass `Natural` to the `mkBackgroundThread` function as well? Also, probably it's not enough to have `stm >= 2.4` because `stm-2.4.5.1` contains very important bugfix for `TBQueue`:

* https://github.com/haskell/stm/issues/16

I even expect from GHC to give you warning about `fromIntegral` usage after you change type from `Int` to `Natural` because we have `-Widentities` warning enabled for `co-log`.

co-log/src/Colog/Concurrent.hs
```diff
@@ -0,0 +1,129 @@
...
+-- loggers in that thread.
+--
+-- You may extend the logger and entire logger will be run
+-- on the background thread. It is not a case with
+-- 'forkBackgroundLogger'.
+
+-- | Create a background worker with a given capacity.
+-- If capacity is reached, then the thread that tries to
+-- write logs will be blocked.
+mkBackgroundThread :: Int -> IO (BackgroundWorker (IO ()))
```
ChShersh: Input message as `IO ()` looks interesting 🤔 I wonder how useful such simple `LogAction` could be:

```haskell
idLogAction :: LogAction m (m ())
idLogAction = LogAction id
```

co-log/src/Colog/Concurrent.hs
```diff
@@ -0,0 +1,129 @@
...
+mkBackgroundThread capacity = do
+  queue <- newTBQueueIO (fromIntegral capacity)
+  tid <- forkIO $ forever $ join $ atomically $ readTBQueue queue
+  pure $ BackgroundWorker tid (writeTBQueue queue)
+
+-- | Run logger action asynchronously in the worker thread.
+-- Logger is executed in the other thread entirely, so if
+-- logger takes any thread related context it will be
+-- read from the other thread.
+runInBackgroundThread :: BackgroundWorker (IO ()) -> LogAction IO msg -> LogAction IO msg
```
ChShersh: It's possible to have the result type of `runInBackgroundThread` function to be `MonadIO m => LogAction m msg` but I wonder how useful it can be. Whole current API is built on top of the plain `IO` and the only sensible generalisation that I know involves `MonadUnliftIO` but I'm not sure it's a good idea to bring `MonadUnliftIO` now.

co-log/src/Colog/Concurrent.hs
```diff
@@ -0,0 +1,129 @@
...
+--
+-- You may extend the logger and but all extensions will be run
+-- in the user thread, only @logger1@ is run in background.
+-- It is not a case with 'runInBackgroundThread'.
+
+-- | Creates background logger with given @Capacity@,
+-- takes a 'LogAction' that should describe how to write
+-- logs.
+--
+-- @capability@ - parameter tells how many in flight messages
```
ChShersh: This variable called `capacity` in multiple places, so probably it's a good idea to name it `capacity` here as well instead of `capability`.

co-log/src/Colog/Concurrent.hs
```diff
@@ -0,0 +1,129 @@
...
+-- are allowed, if that value is reached then user's thread
+-- that emits logs will be blocked until any message will be
+-- written.
```
ChShersh: I think alternative strategy is to drop all log messages after reaching the limit. But I agree with making current behavior the default one. Probably in production it's more important to lose some performance and hang for a while rather than losing important logs about what is actually going on within the running system.

co-log/src/Colog/Concurrent.hs
```diff
@@ -0,0 +1,129 @@
...
+--
+-- N.B. The 'LogAction' will be run in the background
+-- thread so that logger should not add any thread specific
+-- context to the message.
+forkBackgroundLogger :: Int -> LogAction IO msg -> IO (BackgroundWorker msg)
+forkBackgroundLogger capacity logAction = do
+  queue <- newTBQueueIO (fromIntegral capacity)
+  tid <- forkIO $ forever $ do
+    msg <- atomically $ readTBQueue queue
+    unLogAction logAction msg
```
ChShersh: Btw, just general question. Would you appreciate some fancy operator for unwrapping `LogAction` newtype? So instead of

```haskell
unLogAction logAction msg
```

you could write something like:

```haskell
logAction <| msg
```

Or probably have shorter alias for `unLogAction` to partially apply `logAction` in more convenient way?

```haskell
toLogFun logAction :: msg -> m ()
```

co-log/src/Colog/Concurrent.hs
qnikst: It's up to you. As far as I understand `stm-2.5.0.0` will cut many older GHC's, in the case if you want to support 3 major versions then it's not an option. And I find it's easer to keep the same interface for all versions then doing CPP magic to provide the best option for each version.

co-log/src/Colog/Concurrent.hs
qnikst: I'm not very good at CT to find what that object will be and derive a use-case for that based on that knowledge. Right now I don't see how it can be used :/

co-log/src/Colog/Concurrent.hs
qnikst: It can be useful, because that way you may provide useful capabilities there. But I'm uneasy with that change, currently, there are a number of different approaches that allow such generalizations:

1. MonadUnliftIO and unliftio ecosystem in general
2. monad-control that is still widely used
3. MTL-heavy libraries that are using MTL as final tagless system
4. a family of the effects systems.
5. direct handle approach

Making a preference to the one system other the others may be a bad thing, as it increases dependencies set and makes people used to other systems uneasy when using your library.

And `IO` on it's own is very powerful and composable type, in general with any of the approaches mentioned above you can write `LogAction IO msg` and "unlift" your system to those. So basically you need `unlift :: SomeConstraints m => LogAction m msg -> LogAction IO msg` that will be called in the parent thread.

co-log/src/Colog/Concurrent.hs
qnikst: thanks!

co-log/src/Colog/Concurrent.hs
qnikst: I think I'll add some docs and expose more internal so users will be able to define such strategies on their own.  I don't like dropping messages, it should be some very special use-case. If you have too many logs so you can't dump them, it means that your environment has some severe problems that should be solved anyway.

co-log/src/Colog/Concurrent.hs
qnikst: that may be useful, but in internals, I like to use the full name. So it was not difficult for me to write `unLogAction`

qnikst:
@ChShersh, please take another look, I've reworked my branch, fixed few concurrency issues and improved documentation.

co-log/src/Colog/Concurrent.hs
ChShersh: Okay, I can just add `liftLogIO :: MonadIO m => LogAction IO msg -> LogAction m msg` to `co-log-core` to make life of users easier. So they have some way accessible without extra dependencies introduced.

PR #50, opened 2018-10-03
Adds liftLogIO function

### LINKED ISSUE #48: Add `liftLogIO` to `Colog.Core.IO` module:
This function:

```haskell
liftLogIO :: MonadIO m => LogAction IO msg -> LogAction m msg
```

This might be useful.

cronokirby:
Resolves #48.

As per my comment in that issue, this adds `liftLogIO` right next to `liftLogAction` in `Colog.Monad`.

I wasn't sure if this would need a version bump, so I added the changes under "0.1.0" otherwise this addition can go in the next minor.

### DIFF
--- a/co-log/src/Colog/Monad.hs
+++ b/co-log/src/Colog/Monad.hs
@@ -13,7 +13,6 @@ module Colog.Monad
        ) where
 
 import Control.Monad.Reader (MonadReader (..), ReaderT)
-import Control.Monad.IO.Class (MonadIO, liftIO)
 import Control.Monad.Trans.Class (MonadTrans (..))
 
 import Colog.Core (HasLog (..), LogAction (..), overLogAction)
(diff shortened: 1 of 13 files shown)


co-log/src/Colog/Monad.hs
```diff
@@ -65,6 +66,9 @@ withLog = local . overLogAction
 liftLogAction :: (Monad m, MonadTrans t) => LogAction m msg -> LogAction (t m) msg
 liftLogAction (LogAction action) = LogAction (lift . action)
 
+liftLogIO :: MonadIO m => LogAction IO msg -> LogAction m msg
+liftLogIO (LogAction action) = LogAction (liftIO . action)
```
ChShersh: @cronokirby Thanks for your contribution! The reason for this function to be in `co-log-core` package is because it doesn't actually brings new dependencies. `MonadIO` and `liftIO` are already in `base` package. So it's completely okay to put this function to `co-log-core` package (and bump version of `co-log-core` to `0.1.1`). `liftLogAction` on the hand requires `transformers` package because of `MonadTrans` constraint

cronokirby: Ah, yeah that makes sense

PR #51, opened 2018-10-04
[#46] Move log actions that work with String to Colog.Core.IO

### LINKED ISSUE #46: Move log actions that work with `String` from `co-log` to `co-log-core` package
I would like to move the following functions to the `Colog.Core.IO` module:

https://github.com/kowainik/co-log/blob/695a27fe0111918d2e810c9170a2388015595fd9/co-log/src/Colog/Actions.hs#L32-L52

They use only types from the `base` package so they won't introduce extra dependencies. But they are still useful. And they can be used as an example of very basic and default `LogAction` implementations. Also, later they can be used for testing with `doctest`

sphaso:
Resolves #46

### DIFF
--- a/co-log-core/src/Colog/Core/IO.hs
+++ b/co-log-core/src/Colog/Core/IO.hs
@@ -1,0 +1,33 @@
+module Colog.Core.IO 
+    ( logStringStdout
+    , logStringStderr
+    , logStringHandle
+    , withLogStringFile
+    ) where
+
+import Control.Monad.IO.Class (MonadIO, liftIO)
+import System.IO (Handle, IOMode( AppendMode ), hPutStrLn, stderr, withFile)
+import Colog.Core.Action (LogAction (..))
+
+
+{- | Action that prints 'String' to stdout. -}
+logStringStdout :: MonadIO m => LogAction m String
+logStringStdout = LogAction (liftIO . putStrLn)
+
+{- | Action that prints 'String' to stderr. -}
+logStringStderr :: MonadIO m => LogAction m String
+logStringStderr = logStringHandle stderr
+
+{- | Action that prints 'String' to 'Handle'. -}
+logStringHandle :: MonadIO m => Handle -> LogAction m String
+logStringHandle handle = LogAction $ liftIO . hPutStrLn handle
+
+{- | Action that prints 'String' to file. Instead of returning 'LogAction' it's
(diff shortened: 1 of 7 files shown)


co-log-core/src/Colog/Core/IO.hs
```diff
@@ -0,0 +1,37 @@
...
+    ) where
+
+import Control.Monad.IO.Class (MonadIO, liftIO)
+import System.IO (Handle, IOMode( AppendMode ), hPutStrLn, stderr, stdout, withFile)
+import Colog.Core.Action (LogAction (..))
+
+
+----------------------------------------------------------------------------
+-- String
+----------------------------------------------------------------------------
```
ChShersh: I think that this comment is redundant, since this module has only `String` section

sphaso:
@ChShersh thank you for your review, I added the commit to the Changelog and bumped the versions in the `.cabal`. files.
By looking at the changes again I realized that changing

`logStringStdout = LogAction putStrLn` into `logStringStdout = logStringHandle stdout` might have been a mistake on my part, since the new version won't add a newline.

The reason I did this was to overcome this error:
```
../co-log/co-log-core/src/Colog/Core/IO.hs:15:19: error:
        • Couldn't match type ‘m’ with ‘IO’
          ‘m’ is a rigid type variable bound by
            the type signature for:
              logStringStdout :: forall (m :: * -> *).
                                 MonadIO m =>
                                 LogAction m String
            at src/Colog/Core/IO.hs:14:1-50
          Expected type: LogAction m String
            Actual type: LogAction IO String
        • In the expression: LogAction putStrLn
          In an equation for ‘logStringStdout’:
              logStringStdout = LogAction putStrLn
        • Relevant bindings include
            logStringStdout :: LogAction m String
              (bound at src/Colog/Core/IO.hs:15:1)
       |
    15 | logStringStdout = LogAction putStrLn
```
could it be related to the fact that `MonadIO` is now directly from `Control.Monad.IO.Class` and not from `Relude`?

sphaso:
I think that was indeed the point. That function is now:
`logStringStdout = LogAction (liftIO . putStrLn)`

Is this OK?

ChShersh:
@sphaso 

> logStringStdout = LogAction (liftIO . putStrLn)

Yes, this is ok. In `relude` function `putStrLn` is lifted to `MonadIO`, so you need to add `liftIO` (as you already did).

PR #52, opened 2018-10-04
[#39] Build with GHC 8.2.2 and GHC 8.6.1 on CI

### LINKED ISSUE #39: Build with GHC 8.2.2 and GHC 8.6.1 on CI
Since the library is announced, it's a good idea to test 3 latest major versions on CI.

brandonhamilton:
Closes #39

### DIFF
--- a/co-log-core/co-log-core.cabal
+++ b/co-log-core/co-log-core.cabal
@@ -28,7 +28,7 @@ library
                            Colog.Core.Class
                            Colog.Core.Severity
 
-  build-depends:       base >= 4.10 && < 5
+  build-depends:       base >= 4.10 && < 4.13
 
   ghc-options:         -Wall
                        -Wincomplete-uni-patterns
--- a/co-log-core/co-log-core.cabal
+++ b/co-log-core/co-log-core.cabal
@@ -28,7 +28,7 @@ library
                            Colog.Core.Class
                            Colog.Core.Severity
 
-  build-depends:       base >= 4.11 && < 5
+  build-depends:       base >= 4.10 && < 5
 
   ghc-options:         -Wall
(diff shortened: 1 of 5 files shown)


co-log-core/co-log-core.cabal
```diff
@@ -28,7 +28,7 @@ library
                            Colog.Core.Class
                            Colog.Core.Severity
 
-  build-depends:       base >= 4.11 && < 5
+  build-depends:       base >= 4.10 && < 5
```
vrom911: The point to make it `< 4.13` is that when the new `base` version will be released we still can guarantee that this version of our library will work as expected. If we specify `< 5` this would mean that the breaking changes of `base-4.13.*` or later can break this particular version of our library and we will need to make a revision on Hackage to tighten the `base` bounds to `< 4.13`. I hope this explanation makes sense.

brandonhamilton: Yes, that makes sense 👍

brandonhamilton:
Aha - there is now a build issue on 8.2.2

ChShersh:
@brandonhamilton Sorry, turned out I was wrong (so I've deleted my comment). The old interface was only in the GHC 8.0.2 and the new one appeared in GHC 8.2.2. So this is not actually a reason of error...

ChShersh:
@brandonhamilton Okay, looks like GHC 8.2.2 has total disrespect to `forall` in `MessageField` constructor. Writing instance like this solves the issue:

```haskell
    fromLabel field = TM.WrapTypeable $ MessageField @_ @fieldName field
```

So I think you still can use `-XCPP` and resolve this particular issue. But then you will observe the compilation error with `README.lhs`...

PR #54, opened 2018-10-05
Run doctests for co-log-core

### LINKED ISSUE #47: Add `doctest` tests to `co-log-core` package
Blocked by #46 

It's a great idea to test all functions in `Colog.Core` package using `doctest`. After #46 is implemented, `logStringStdout` action can be used as testing example. This will improve documentation significantly.

tfausak:
Fixes #47.

### DIFF
--- a/co-log-core/src/Colog/Core/Action.hs
+++ b/co-log-core/src/Colog/Core/Action.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE InstanceSigs #-}
 {-# LANGUAGE ViewPatterns #-}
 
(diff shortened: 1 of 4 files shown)


co-log-core/src/Colog/Core/Action.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE InstanceSigs #-}
```
ChShersh: This extension is enabled by default for whole project. It's very useful! So I think that instead of adding it to every manually, you can just write:

```haskell
main = Doctest.doctest [ "-XInstanceSigs", "src" ]
```

tfausak:
💞 Thanks!

PR #57, opened 2018-10-08
[#49] Adds and exports (<&) and (&>) comibinators for LogAction

### LINKED ISSUE #49: Add fancy operator to pass `msg` to `LogAction m msg`
Currently, in order to pass `msg` to `LogAction m msg` you need to write the following code:

```haskell
unLogAction logAction msg
```

This might not be convenient. It would be much cooler to have some fancy operator to pass logs to actions, like this one:

```haskell
logAction <| msg
```

Maybe `<|` is not the best choice, since it's already a popular operator name. So I'm open to proposal for better names.

cronokirby:
Resolves #49 

Adds 2 operator versions of `unLogAction`. I've given both of them a precedence of 5, so that they're higher than the other combinators, so that things like:
```haskell
f >$< log <& msg
```
have to be written as:
```haskell
(f >$< log) <& msg
```
The infixity should also be > 2 to be higher than the monad operators `>>` and `>>=`, so that
we can do things like:
```haskell
msg &> action >> msg &> action
```

Because something like
```haskell
action <& msg <& msg
```
doesn't make sense in terms of the types, I haven't added infixr or infixl.

### DIFF
--- a/co-log-core/src/Colog/Core/Action.hs
+++ b/co-log-core/src/Colog/Core/Action.hs
@@ -122,13 +122,27 @@ Note that because of the types, something like:
 > action <& msg1 <& msg2
 doesn't make sense. Instead you want:
 > action <& msg1 >> action <& msg2
+
+In addition, because '<&' has higher precedence
+than the other operators in this module,
+the following:
+> f >$< action <& msg
+should be replaced by
+> (f >$< action) <& msg
+
 -}
 infix 5 <&
 (<&) :: LogAction m msg -> msg -> m ()
 (<&) = coerce
 {-# INLINE (<&) #-}
 
--- | A flipped version of '<&'
+{- | A flipped version of '<&'
+
+It shares the same precedence as '<&',
+so make sure to surround lower precedence
+operators in parentheses:
+> (f >$< action) <& msg
+-}
 infix 5 &>
 (&>) :: msg -> LogAction m msg -> m ()
 (&>) = flip unLogAction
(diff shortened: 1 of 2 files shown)


co-log-core/src/Colog/Core/Action.hs
```diff
@@ -112,6 +114,23 @@ instance Applicative m => Monoid (LogAction m a) where
...
+Note that because of the types, something like:
+> action <& msg1 <& msg2
+doesn't make sense. Instead you want:
+> action <& msg1 >> action <& msg2
+-}
+infix 5 <&
+(<&) :: LogAction m msg -> msg -> m ()
+(<&) = unLogAction
+
+-- | A flipped version of <&
```
vrom911: To have it showed as the link to the `<&` function in the haddock documentation you should put it into quotes: `'<&'`

vrom911:
Oh, forgot to mention, can you also update the `CHANGELOG`?

co-log-core/src/Colog/Core/Action.hs
cronokirby: Right

co-log-core/src/Colog/Core/Action.hs
```diff
@@ -7,6 +7,8 @@
 module Colog.Core.Action
        ( -- * Core type and instances
          LogAction (..)
+        , (<&)
+        , (&>)
```
ChShersh: Here you have 8 spaces, while it should be 7 according to this module code style

cronokirby: Hmm, perhaps we should look into adding a stylish haskell rule for this if we can?

vrom911: Unfortunately, `stylish-haskell` doesn't have this configurations :(

ChShersh: Yep, here is an open issue:

* https://github.com/jaspervdj/stylish-haskell/issues/220

PR #61, opened 2018-10-12
Create doctests for `Colog.Core.Action`

### LINKED ISSUE #58: Add doctest for every function in the `Colog.Core.Action` module
We have nice operators `&>` and `<&` to pass messages to loggers. Very useful for doctesting!

cronokirby:
Resolves #58.

Expands the documentation for functions, by adding documentation tests.

There were cases in which a function was heavily documented with an example that wasn't tested, e.g.
using a combinator to log both to files and stdout; in those cases I didn't change them. In the cases
where a combinator / function was just a flipped version or synonym of another, I also didn't document any more, as I felt this was redundant.

I also noticed that `conquer` had the same implementation as `mempty`, so defined the former
via the latter to avoid code duplication.


co-log-core/src/Colog/Core/Action.hs
```diff
@@ -210,7 +210,12 @@ cmap :: (a -> b) -> LogAction m b -> LogAction m a
 cmap f (LogAction action) = LogAction (action . f)
 {-# INLINE cmap #-}
 
--- | Operator version of 'cmap'.
+{- | Operator version of 'cmap'.
+
+>>> logString = LogAction putStrLn
```
ChShersh: I think it's better to use `logStringStdout` from `Colog.Core.IO` module

co-log-core/src/Colog/Core/Action.hs
```diff
@@ -270,30 +281,62 @@ cmapM :: Monad m => (a -> m b) -> LogAction m b -> LogAction m a
 cmapM f (LogAction action) = LogAction (f >=> action)
 {-# INLINE cmapM #-}
 
--- | @divide@ combinator from @Divisible@ type class.
+{- | @divide@ combinator from @Divisible@ type class.
+
+>>> logString = LogAction putStrLn
+>>> logInt = LogAction print
```
ChShersh: Btw, `LogAction print` looks like a good addition to the `Colog.Core.IO` module

co-log-core/src/Colog/Core/Action.hs
```diff
@@ -270,30 +281,62 @@ cmapM :: Monad m => (a -> m b) -> LogAction m b -> LogAction m a
...
+3
+-}
 divide :: (Applicative m) => (a -> (b, c)) -> LogAction m b -> LogAction m c -> LogAction m a
 divide f (LogAction actionB) (LogAction actionC) = LogAction $ \(f -> (b, c)) ->
     actionB b *> actionC c
 
--- | @conquer@ combinator from @Divisible@ type class.
+{- | @conquer@ combinator from @Divisible@ type class.
+
+Concretely, this is a logAction that does nothing:
```
ChShersh: Instead of `logAction` it's better to write `'LogAction'` to convert it into link

co-log-core/src/Colog/Core/Action.hs
cronokirby: Should I go ahead and add this in this pr, or open should we open an issue?

co-log-core/src/Colog/Core/Action.hs
cronokirby: This would require a cyclic dependency, since Core.IO depends on Core.Action. Or are you suggesting to reuse the name from Core.IO for clarity?

ChShersh: @cronokirby cyclic dependency is not a problem, because `doctest` has it's own `$setup` comment to import other modules. See this module in `relude` for example:

https://github.com/kowainik/relude/blob/9e3aa751851b5a4d4301119c9053275e520fb977/src/Relude/String/Conversion.hs#L62-L64

Even if `Relude` imports `Relude.String.Conversion`, it's still okay to import `Relude` in the comments.

co-log-core/src/Colog/Core/Action.hs
ChShersh: @cronokirby I think separate issue is better 👍

co-log-core/src/Colog/Core/Action.hs
cronokirby: Good to know!

PR #62, opened 2018-10-13
[#20] Add logger rotation action

### LINKED ISSUE #20: Add logger rotation action

sphaso:
(will) Resolve #20 

I'm currently trying to solve the compilation error, I guess I should combine `putStrLn` somewhere but the type tetris is failing me. I opened the PR anyway to check if you like the approach or if you have any other remark, please do tell.

### DIFF
--- a/co-log/src/Colog/Actions.hs
+++ b/co-log/src/Colog/Actions.hs
@@ -14,7 +14,6 @@ module Colog.Actions
        ) where
 
 import Colog.Core.Action (LogAction (..))
-
 import qualified Data.ByteString.Char8 as BS
 import qualified Data.Text.IO as TIO
 
@@ -58,43 +57,3 @@ logTextHandle handle = LogAction $ liftIO . TIO.hPutStrLn handle
 withLogTextFile :: MonadIO m => FilePath -> (LogAction m Text -> IO r) -> IO r
 withLogTextFile path action = withFile path AppendMode $ action . logTextHandle
 
-----------------------------------------------------------------------------
--- Logger rotation
-----------------------------------------------------------------------------
-
-{-
-data Limit = LimitTo Natural | Unlimited
-
-{- | Logger rotation action. Takes name of the logging file @file.foo@. Always
-writes new logs to file named @file.foo@ (given file name, also called as /hot log/).
-
-* If the size of the file exceeds given limit for file sizes then this action
-  renames @file.foo@ to @file.foo.(n + 1)@ (where @n@ is the number of latest
-  renamed file).
-* If the number of files on the filesystem is bigger than the files number limit
-  then the given @FilePath -> IO ()@ action is called on the oldest file. As
-  simple solution, you can pass @removeFile@ function to delete old files but
-  you can also pass some archiving function if you don't want to loose old logs.
--}
-withLogRotation
-    :: forall msg m .
-       MonadIO m
-    => Limit  -- TODO: use 'named' library here to distinguish limits?
-    -- ^ Max allowed file size in bytes
-    -> Limit
-    -- ^ Max allowed number of files to have
-    -> FilePath
-    -- ^ File path to log
-    -> (FilePath -> IO ())
-    -- ^ What to do with old files; pass @removeFile@ here for deletion
-    -> (Handle -> LogAction m msg)
-    -- ^ Action that writes to file handle
-    -> (LogAction m msg -> IO r)
-    -- ^ Continuation action
-    -> IO r
-withLogRotation sizeLimit filesLimit path cleanup mkAction cont = cont rotationAction
-  where
-    rotationAction :: LogAction m msg
-    rotationAction = LogAction $ \msg -> do
-        withFile path AppendMode writeFileLoop
--}
--- a/co-log/src/Colog/Actions.hs
+++ b/co-log/src/Colog/Actions.hs
@@ -14,6 +14,7 @@ module Colog.Actions
        ) where
 
 import Colog.Core.Action (LogAction (..))
+
 import qualified Data.ByteString.Char8 as BS
 import qualified Data.Text.IO as TIO
 
--- a/co-log/src/Colog/Rotation.hs
+++ b/co-log/src/Colog/Rotation.hs
@@ -1,14 +1,20 @@
+-- |
+-- This functionality is not to be considered stable
+-- or ready for production use. While we enourage you
+-- to try it out and report bugs, we cannot assure you
+-- that everything will work as advertised :)
+
 module Colog.Rotation
-       (
-         Limit(..)
+       ( Limit(..)
        , withLogRotation
        ) where
 
 import Data.Maybe (mapMaybe)
 import Data.Semigroup (Max (..))
+import System.FilePath.Posix ((<.>))
 import System.IO (hFileSize)
 
-import Colog.Core.Action (LogAction (..))
+import Colog.Core.Action (LogAction (..), (<&))
 
 import qualified System.Directory as D
 import qualified System.FilePath.Posix as POS
@@ -54,26 +60,27 @@ withLogRotation sizeLimit filesLimit path cleanup mkAction cont = do
     rotationAction refHandle
       = LogAction $ \msg -> do
         handle <- liftIO $ readIORef refHandle
-        unLogAction (mkAction handle) msg
+        mkAction handle <& msg
 
         whenM
           (liftIO $ isFileSizeLimitReached sizeLimit handle)
           (cleanupAndRotate refHandle)
+
     cleanupAndRotate :: IORef Handle -> m ()
     cleanupAndRotate refHandle = liftIO $ do
-      h <- readIORef refHandle
-      hClose h
+      readIORef refHandle >>= hClose
       maxN <- maxFileIndex path
-      renameFileToNumber (succ maxN) path
+      renameFileToNumber (maxN + 1) path
       oldFiles <- getOldFiles filesLimit path
       mapM_ cleanup oldFiles
       newHandle <- openFile path AppendMode
-      modifyIORef' refHandle (const newHandle)
+      writeIORef refHandle newHandle
 
 isLimitedBy :: Integer -> Limit -> Bool
 isLimitedBy _ Unlimited = False
-isLimitedBy size (LimitTo limit) | size <= 0 = False
-                                 | otherwise = limit > (fromInteger size :: Natural)
+isLimitedBy size (LimitTo limit)
+  | size <= 0 = False
+  | otherwise = limit > (fromInteger size :: Natural)
 
 isFileSizeLimitReached :: Limit -> Handle -> IO Bool
 isFileSizeLimitReached limit handle = do
@@ -91,7 +98,7 @@ maxFileIndex path = do
 
 -- given number 4 and path `node.log` renames file `node.log` to `node.log.4`
 renameFileToNumber :: Int -> FilePath -> IO ()
-renameFileToNumber n path = D.renameFile path (path POS.<.> show n)
+renameFileToNumber n path = D.renameFile path (path <.> show n)
 
 -- if you give it name like `node.log.4` then it returns `Just 4`
 logFileIndex :: FilePath -> Maybe Int
@@ -109,4 +116,3 @@ getOldFiles limit path = do
     isOldFile maxN n = case limit of
                          Unlimited -> False
                          LimitTo l -> n < toInteger maxN - toInteger l
-
--- a/co-log/src/Colog/Rotation.hs
+++ b/co-log/src/Colog/Rotation.hs
@@ -1,0 +1,112 @@
+module Colog.Rotation
+       (
+         Limit(..)
+       , withLogRotation
+       ) where
+
+import Data.Maybe (mapMaybe)
+import Data.Semigroup (Max (..))
+import System.IO (hFileSize)
+
+import Colog.Core.Action (LogAction (..))
+
+import qualified System.Directory as D
+import qualified System.FilePath.Posix as POS
+
+
+data Limit = LimitTo Natural | Unlimited deriving (Eq, Ord)
+
+{- | Logger rotation action. Takes name of the logging file @file.foo@. Always
+writes new logs to file named @file.foo@ (given file name, also called as /hot log/).
+
+* If the size of the file exceeds given limit for file sizes then this action
+  renames @file.foo@ to @file.foo.(n + 1)@ (where @n@ is the number of latest
+  renamed file).
+* If the number of files on the filesystem is bigger than the files number limit
+  then the given @FilePath -> IO ()@ action is called on the oldest file. As
+  simple solution, you can pass @removeFile@ function to delete old files but
+  you can also pass some archiving function if you don't want to lose old logs.
+-}
+withLogRotation
+    :: forall r msg m .
+       MonadIO m
+    => Limit
+    -- ^ Max allowed file size in bytes
+    -> Limit
+    -- ^ Max allowed number of files to have
+    -> FilePath
+    -- ^ File path to log
+    -> (FilePath -> IO ())
+    -- ^ What to do with old files; pass @removeFile@ here for deletion
+    -> (Handle -> LogAction m msg)
+    -- ^ Action that writes to file handle
+    -> (LogAction m msg -> IO r)
+    -- ^ Continuation action
+    -> IO r
+withLogRotation sizeLimit filesLimit path cleanup mkAction cont = do
+    -- TODO: figure out how to use bracket to safely manage
+    -- possible exceptions
+    handle <- openFile path AppendMode
+    handleRef <- newIORef handle
+    cont $ rotationAction handleRef
+  where
+    rotationAction :: IORef Handle -> LogAction m msg
+    rotationAction refHandle
+      = LogAction $ \msg -> do
+        handle <- liftIO $ readIORef refHandle
+        unLogAction (mkAction handle) msg
+
+        whenM
+          (liftIO $ isFileSizeLimitReached sizeLimit handle)
+          (cleanupAndRotate refHandle)
+    cleanupAndRotate :: IORef Handle -> m ()
+    cleanupAndRotate refHandle = liftIO $ do
+      h <- readIORef refHandle
+      hClose h
+      maxN <- maxFileIndex path
+      renameFileToNumber (succ maxN) path
+      oldFiles <- getOldFiles filesLimit path
+      mapM_ cleanup oldFiles
+      newHandle <- openFile path AppendMode
+      modifyIORef' refHandle (const newHandle)
+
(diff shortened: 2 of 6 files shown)


co-log/src/Colog/Actions.hs
```diff
@@ -92,9 +102,37 @@ withLogRotation
...
+    mapM_ (D.renameFile path) toRename
+
+    cont rotationAction
   where
     rotationAction :: LogAction m msg
-    rotationAction = LogAction $ \msg -> do
-        withFile path AppendMode writeFileLoop
--}
+    rotationAction = liftLogIO $ LogAction $ (\msg ->
+        withFile path AppendMode (flip (unLogAction . mkAction) msg))
```
ChShersh: You got the idea right. But returns `rotationAction` is not valid, unfortunately. You return this action and it's supposed to do all the dirty work. So on every call to `LogAction` this function should output line to the file, check resulted length of the file and rename files if needed. Currently you're performing renaming only once, at the beginning of `withLogRotation` while in reality this should be done on every call to `rotationAction`.

The difficulty in implementing this function is checking the size of the file each time, keep outputting to `Handle` instead of reopening file each time and use new `Handle` each time the size of the file is bigger than the given `Limit`

ChShersh: > Would we expect the user to have a partially applied withLogRotation and pass cont each time they want to log?

The use-case is that you call `withLogRotation` only once in your program to create properly configured `LogAction` and then you use only that action. Compare with more simple example: `withFile`. It's more efficient to open file only once and use `Handle` to write output instead of calling `appendFile` on every message. So you open file only once in your program, and then you use given action to output the result. Same with `withLogRotation`. You specify configuration parameters and then you can use single action to output messages, and this action handles file renaming and size check for you.

co-log/src/Colog/Actions.hs
```diff
@@ -62,8 +67,13 @@ withLogTextFile path action = withFile path AppendMode $ action . logTextHandle
 -- Logger rotation
 ----------------------------------------------------------------------------
 
-{-
-data Limit = LimitTo Natural | Unlimited
+
+data Limit = LimitTo Natural | Unlimited deriving (Eq)
```
ChShersh: Actually, if you write constructors in this particular order, the derived `Ord` instances will be correct, no need to write it manually

ChShersh:
@sphaso Thank you for continuing to work on this! The issue is actually is pretty difficult, so no easy way to resolve it...

> Do you think msg can be kept polymorphic? I see other withFile examples have it constrained to either String, ByteString or Text...

I think that `msg` can be kept polymorphic. The idea is that `withLogRotation` action will be called like this:

```haskell
main :: IO ()
main = withLogRotation (Limit 3) (Limit 5) "node.log" removeFile logByteStringHandle $ \action ->
    ... the rest of the program ...
```

And instead of passing `logByteStringHandle` I can pass `logTextHandle` or `logStringHandle`. So there actually shouldn't be any logic related to the type of `msg` rather than just calling given action on the `msg`. You can write `msg` to `Handle` and get size of `Handle` after outputting message. So no need to calculate size of `msg`.

ChShersh:
@sphaso I have an idea on how to implement this. I think that this can be done only using `IORef Handle` and manually working with `open` and `hClose` functions carefully cleaning up resources. I don't see way to implement this using `withFile` function (though it would be really nice to do so). 

And I think that the issue should be separated into smaller ones... Like:

1. Implement `Limit` data type, instances for this data type and unit tests for `Ord` instance.
2. Implement function that walks over files with lower numbers and applies given cleaning function to them.
3. Implement function that writes to `Handle` and checks the size of the `Handle` and calls previous function if case if file exceeds limit.
4. Combine everything into `withLogRotation` function.

sphaso:
@ChShersh Hi, I've arrived to a similar conclusion (not being able to use `withFile`, but using `open` instead). I'll implement the steps you mentioned, although I think 2 and 3 might be merged into a single function? (that's what I did in the code).     
Right now I'm a bit stuck with the monad stack types, but I feel very close to a conclusion!

sphaso:
@ChShersh thanks to the help of @alpacaaa I was able to have a compiling version. I'm not sure what you mean by
> Implement Limit data type, instances for this data type and unit tests for Ord instance.

Do you mean that we should have `data Limit a = Limit a | Unlimited` instead of `Limit Natural`? Does it need to have its own class?

Regarding point 2 and 3, as you can tell from the code, I make both checks every time, since I supposed both conditions need to pass before writing to the file (both size and file numbers need to be below the limit).

ChShersh:
@sphaso 

>

I meant exactly what you did, so good job 👍 

```haskell
data Limit = LimitTo Natural | Unlimited deriving (Eq, Ord)
```

I've checked the derived instance and it works as expected 👍

screenshot 2018-10-29 at 12 20 27 pm

sphaso:
@ChShersh work done since last commit:
- put `ScopedTypeVariables` in cabal (and removed it from two other modules in `co-log`)
- exported `Limit` and `withLogRotation` from Action module
- implemented the functions you suggested and modified the code accordingly, it looks much cleaner!
- removed the `IORef` from `withLogRotation` parameters (we had added it to the parameters because we though that by creating it into the body it would make a new one each time the `cont` would be applied)
- used `bracket` (very nice addition too!)
Now I'm battling against the usual error: `Couldn't match type ‘m’ with ‘IO’`
but I imagine you might want to change more code, so I won't focus too much on that.

As an aside, I wanted to thank you guys for making your projects so beginner-friendly, especially for the hacktoberfest. It's been a blast and I hope to continue contributing in the future. My Haskell knowledge sure improved a lot!

co-log/src/Colog/Actions.hs
```diff
@@ -92,9 +102,64 @@ withLogRotation
...
-        withFile path AppendMode writeFileLoop
--}
+    rotationAction refHandle
+      = LogAction $ \msg -> do
+          oldFiles <- getOldFiles filesLimit path
+          mapM_ (liftIO . cleanup) oldFiles
+          handle <- readIORef refHandle
+          sizeLimitReached <- isFileSizeLimitReached sizeLimit handle
+
+          if sizeLimitReached then do
```
ChShersh: Actually, the logic can be slightly simpler. It's okay to always write `msg` to the existing `Handle` and perform renaming only after message is written. Like this:

```haskell
outputMsg msg handle
whenM (isFileSizeLimitReached sizeLimit handle) cleanupAndRotate
```

Also, you need to cleanup only when you perform logger rotation. Don't need to cleanup every time on each message.

co-log/src/Colog/Actions.hs
```diff
@@ -92,9 +102,64 @@ withLogRotation
...
+renameFileToNumber :: Int -> FilePath -> IO ()
+renameFileToNumber n path = D.renameFile path (path <> "." <> show n)
+
+-- if you give it name like `node.log.4` then it returns `Just 4`
+logFileIndex :: FilePath -> Maybe Int
+logFileIndex path = readMaybe =<< (tail <$> (nonEmpty $ POS.takeExtension path))
+
+-- creates list of files with indices who are older on given Limit than the latest one
+getOldFiles :: Limit -> FilePath -> IO [FilePath]
+getOldFiles limit path = do
```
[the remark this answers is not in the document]
sphaso: Thank you, this is very informative! I guess it was fair to assume I should have understood this from the start

sphaso:
@ChShersh sorry for the sloppy style. This time I used `stylish-haskell` and double checked the style guidelines, so everything should be properly aligned and ordered.

I think I implemented all the changes you mentioned. The code looks so much cleaner now :) 
I have some questions:

**1. type inference**
Why does this work
```Haskell
    cleanupAndRotate refHandle = liftIO $ do
      h <- readIORef refHandle
      hClose h
```
but `hClose <$> readIOref refHandle` gives an error for an ambiguos MonadIO?
```
Could not deduce (MonadIO m0) arising from a use of ‘hClose’
      from the context: MonadIO m
```
It took me a while to figure out that just breaking the function could work, but it was sheer intuition, not knowledge.

**2. algebra**
There's two points where I check if `Limit` is `Unlimited` or `LimitTo n`, does this warrant having a `Functor` instance of `Limit`?

ChShersh:
@sphaso 

> Why does this work .. but ... gives an error for an ambiguos MonadIO?

I actually don't know. Depends on the context. But `hClose <$> readIOref refHandle` is not the equivalent of

```haskell
    cleanupAndRotate refHandle = liftIO $ do
      h <- readIORef refHandle
      hClose h
```

it should be `readIORef refHandle >>= hClose`. I think because of that type inference got broken.

> 2. algebra
> There's two points where I check if Limit is Unlimited or LimitTo n, does this warrant having a Functor > instance of Limit?

Well, `Limit` data type is not polymorphic so having `Functor` instance for it won't help. If you pattern-match on `Limit` constructors multiple types, you can implement function:

```haskell
limit :: a -> (Natural -> a) -> Limit -> a
```

which is called `Limit`-eliminator (similar to `maybe :: b -> (a -> b) -> Maybe a -> b` which is called `Maybe`-eliminator).

If having such `limit` function helps you then it's okay 🙂

co-log/src/Colog/Rotation.hs
```diff
@@ -0,0 +1,112 @@
...
+withLogRotation sizeLimit filesLimit path cleanup mkAction cont = do
+    -- TODO: figure out how to use bracket to safely manage
+    -- possible exceptions
+    handle <- openFile path AppendMode
+    handleRef <- newIORef handle
+    cont $ rotationAction handleRef
+  where
+    rotationAction :: IORef Handle -> LogAction m msg
+    rotationAction refHandle
+      = LogAction $ \msg -> do
```
ChShersh: I think that this function can be written slightly shorter. No need to move on separate line:

```haskell
rotationAction refHandle = LogAction $ \msg -> do
    handle <- ...
```

co-log/src/Colog/Rotation.hs
```diff
@@ -0,0 +1,112 @@
...
+        handle <- liftIO $ readIORef refHandle
+        unLogAction (mkAction handle) msg
+
+        whenM
+          (liftIO $ isFileSizeLimitReached sizeLimit handle)
```
ChShersh: You can put `liftIO` in front of `do` in `isFileSizeLimitReached` function to make this `whenM` call look slightly nicer.

co-log/src/Colog/Rotation.hs
```diff
@@ -0,0 +1,112 @@
...
+          (cleanupAndRotate refHandle)
+    cleanupAndRotate :: IORef Handle -> m ()
+    cleanupAndRotate refHandle = liftIO $ do
+      h <- readIORef refHandle
+      hClose h
+      maxN <- maxFileIndex path
+      renameFileToNumber (succ maxN) path
```
ChShersh: `succ` is slower than `+ 1`

```suggestion
      renameFileToNumber (maxN + 1) path
```

co-log/src/Colog/Rotation.hs
```diff
@@ -0,0 +1,112 @@
...
+      oldFiles <- getOldFiles filesLimit path
+      mapM_ cleanup oldFiles
+      newHandle <- openFile path AppendMode
+      modifyIORef' refHandle (const newHandle)
```
ChShersh:
```suggestion
      writeIORef refHandle newHandle
```

co-log/src/Colog/Rotation.hs
```diff
@@ -0,0 +1,112 @@
...
+  let logFiles = filter (== POS.takeBaseName path) files
+  let maxFile = getMax . foldMap Max <$> nonEmpty (mapMaybe logFileIndex logFiles)
+  pure $ fromMaybe 0 maxFile
+
+-- given number 4 and path `node.log` renames file `node.log` to `node.log.4`
+renameFileToNumber :: Int -> FilePath -> IO ()
+renameFileToNumber n path = D.renameFile path (path POS.<.> show n)
+
+-- if you give it name like `node.log.4` then it returns `Just 4`
+logFileIndex :: FilePath -> Maybe Int
```
ChShersh: I feel that code can be simplified if instead of `Maybe Int` you will return `Maybe Natural` here

co-log/src/Colog/Rotation.hs
```diff
@@ -0,0 +1,112 @@
...
+logFileIndex path = nonEmpty (POS.takeExtension path) >>= readMaybe . tail
+
+-- creates list of files with indices who are older on given Limit than the latest one
+getOldFiles :: Limit -> FilePath -> IO [FilePath]
+getOldFiles limit path = do
+    currentMaxN <- maxFileIndex path
+    files <- D.listDirectory (POS.takeDirectory path)
+    let tuple = map (\a -> (a, toInteger <$> logFileIndex a)) files
+    pure $ map fst $ filter (maybe False (isOldFile currentMaxN) . snd) tuple
```
ChShersh: I feel like this two lines of code can be simplified. For example, you can use `mapMaybe` function to get rid of `maybe False` and `filter`

co-log/src/Colog/Rotation.hs
```diff
@@ -0,0 +1,112 @@
...
+  where
+    isOldFile :: Int -> Integer -> Bool
+    isOldFile maxN n = case limit of
+                         Unlimited -> False
+                         LimitTo l -> n < toInteger maxN - toInteger l
+
```
ChShersh: Looks like we need `hintman` here :suspect:

co-log/src/Colog/Rotation.hs
sphaso: This is surprising... any idea why?

co-log/src/Colog/Rotation.hs
sphaso: I would need a function like:
`sequenceC :: (Traversable t, Monad m) => t (a, m b) -> t (m (a, b))`
I could write it myself, but it would complicate the code just to clean these two lines. Any idea?

co-log/src/Colog/Rotation.hs
sphaso: This is a brilliant intuition!

sphaso:
@ChShersh Thanks again for the review. I think I fixed most of comments, I only left 2 where I have some doubts :)

co-log/src/Colog/Rotation.hs
ChShersh: @sphaso `succ` checks for overflow. That's why is slower. You might be surprised by the following behavior as well:

```haskell
ghci> succ (maxBound :: Int)
*** Exception: Prelude.Enum.succ{Int}: tried to take `succ' of maxBound
```

co-log/src/Colog/Rotation.hs
```diff
@@ -0,0 +1,112 @@
...
+-- creates list of files with indices who are older on given Limit than the latest one
+getOldFiles :: Limit -> FilePath -> IO [FilePath]
+getOldFiles limit path = do
+    currentMaxN <- maxFileIndex path
+    files <- D.listDirectory (POS.takeDirectory path)
+    let tuple = map (\a -> (a, toInteger <$> logFileIndex a)) files
+    pure $ map fst $ filter (maybe False (isOldFile currentMaxN) . snd) tuple
+  where
+    isOldFile :: Int -> Integer -> Bool
+    isOldFile maxN n = case limit of
```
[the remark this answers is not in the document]
ChShersh: @sphaso Makes sense. I'm okay with having separate function for that :+1: For me it's hard to perform such kind of math in my head. Thank you for writing the equations!

co-log/src/Colog/Rotation.hs
ChShersh: Okay, it's usually helps to write types of functions to see how it can be simplified :slightly_smiling_face: Usually in such situations Haskell reminds this child game where you have objects of different shapes and you need to put them inside box where holes of those shapes should match :slightly_smiling_face:  

Context:

```haskell
files :: [FilePath]
logFileIndex :: FilePath -> Maybe Natural
isOldFile maxN :: Natural -> Bool
mapMaybe :: (a -> Maybe b) -> [a] -> [b]

<goal> :: [FilePath] -- by `boolean predicate
```

The idea is to implement couple functions:

```haskell
takeFileIndex  :: FilePath -> Maybe (FilePath, Natural)
guardFileIndex :: (FilePath, Natural) -> Maybe FilePath
-- should be quite easy using the existing functions
```

Then you will have:

```haskell
mapMaybe (takeFileIndex >=> guardFileIndex) files :: [FilePath]
```

No need to play with `Bool`, I personally like `mapMaybe` function, it's really powerful.

co-log/src/Colog/Rotation.hs
```diff
@@ -0,0 +1,118 @@
+-- |
+-- This functionality is not to be considered stable
+-- or ready for production use. While we enourage you
+-- to try it out and report bugs, we cannot assure you
+-- that everything will work as advertised :)
```
ChShersh: Very nice text!

co-log/src/Colog/Rotation.hs
```diff
@@ -0,0 +1,118 @@
...
+      oldFiles <- getOldFiles filesLimit path
+      mapM_ cleanup oldFiles
+      newHandle <- openFile path AppendMode
+      writeIORef refHandle newHandle
+
+isLimitedBy :: Integer -> Limit -> Bool
+isLimitedBy _ Unlimited = False
+isLimitedBy size (LimitTo limit)
+  | size <= 0 = False
+  | otherwise = (toInteger limit :: Integer) > size
```
ChShersh: Should work without explicit type annotation, type inference works :+1: 
```suggestion
  | otherwise = toInteger limit > size
```

co-log/src/Colog/Rotation.hs
```diff
@@ -0,0 +1,118 @@
...
+    cleanupAndRotate refHandle = liftIO $ do
+      readIORef refHandle >>= hClose
+      maxN <- maxFileIndex path
+      renameFileToNumber (maxN + 1) path
+      oldFiles <- getOldFiles filesLimit path
+      mapM_ cleanup oldFiles
+      newHandle <- openFile path AppendMode
+      writeIORef refHandle newHandle
+
+isLimitedBy :: Integer -> Limit -> Bool
```
ChShersh: I think, it's a good idea to specify explicitly in the comment for this function whether `isLimitedBy n limit` means `n < limit` or `n <= limit`.

co-log/src/Colog/Rotation.hs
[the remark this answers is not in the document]
sphaso: not a problem, I had to write it down myself :)

co-log/src/Colog/Rotation.hs
sphaso: Wow this is amazing. I'm seeing a pattern in these refactorings: when things look like they can't compose, break them down :)

sphaso:
@ChShersh more refactoring has been done :)

PR #64, opened 2018-10-14
Introduce basic benchmarks of the library.

qnikst:
Introduce basic benchmarks of the log mechanism that allows
testing the overhead of the each feature added.

Example output:
```
Dump 10k messages (in a forked process):
baseline                                           0.001114674s
stdout<string                                      0.005194026s
stdout<text                                        0.004967748s
stdout<bytestring                                  0.002918046s
stderr<bytestring                                  0.01753503s
{stdout<>stderr}<bytestring                        0.017461043s
stdout<format<message                              0.009314225s
stdout<format<message{with callstack}              0.009744045s
stdout<format<message{with callstack:5}            0.018102189s
stdout<format<message{with callstack:50}           0.018013676s
stdout<bytestring<format<message                   0.009570143s
stdout<bytestring<rich-message<upgrade<message     0.134770134s
```
Currently benchmarks are implemented outside of the main
application as I had trouble to make `cabal new-configure` find
the build plan without `--allow-newer`. Also these benchmarks
may introduce dependencies on the other logging libraries in the
future.

At this point of time only the simplest benchmarks were introduced.
This benchmarks are testing sequencial dump of the logs with different
features added atop. Theoretically current benchmarks may be moved
to the application and criterion can be used for measuring time.
However I planned to make more long running benchmarks where criterion
may not shine, and such benchmarks may make CI time much larger for
not purpose.

### DIFF
--- a/co-log-benchmark/LICENSE
+++ b/co-log-benchmark/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2018 Alexander Vershilov
+Copyright (c) Kowainik
 
 Permission is hereby granted, free of charge, to any person obtaining
 a copy of this software and associated documentation files (the
--- a/co-log-benchmark/LICENSE
+++ b/co-log-benchmark/LICENSE
@@ -1,0 +1,20 @@
+Copyright (c) 2018 Alexander Vershilov
+
+Permission is hereby granted, free of charge, to any person obtaining
+a copy of this software and associated documentation files (the
+"Software"), to deal in the Software without restriction, including
+without limitation the rights to use, copy, modify, merge, publish,
+distribute, sublicense, and/or sell copies of the Software, and to
+permit persons to whom the Software is furnished to do so, subject to
+the following conditions:
+
+The above copyright notice and this permission notice shall be included
+in all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
--- a/co-log-benchmark/Main.hs
+++ b/co-log-benchmark/Main.hs
@@ -1,0 +1,103 @@
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings #-}
+module Main
+  ( main
+  ) where
+
+import Data.Time.Clock
+import Colog
+import Control.Monad
+import Control.Exception
+import Data.Foldable
+import System.Environment
+import System.IO
+import System.Process.Typed
+import System.Posix.Process
+import qualified Data.Text.Encoding
+import GHC.Stack
+import Text.Printf
+
+
+-- | List of benchmarks.
+benchs :: [(String, IO ())]
+benchs =
+  [("baseline", run (LogAction (\_ -> pure ())) ["message"::String])
+  ,("stdout<string",
+     let la = logStringStdout 
+     in run la ["message"])
+  ,("stdout<text",
+     let la = logTextStdout
+     in run la ["message"])
+  ,("stdout<bytestring",
+     let la = logByteStringStdout 
+     in run la ["message"])
+  ,("stderr<bytestring",
+     let la = logByteStringStderr 
+     in run la ["message"])
+  ,("{stdout<>stderr}<bytestring",
+     let la = logByteStringStdout <> logByteStringStderr
+     in run la ["message"])
+  ,("stdout<format<message",
+     let la = cmap fmtMessage logTextStdout
+     in run la [Message D emptyCallStack "message"])
+  ,("stdout<format<message{with callstack}",
+     let la = cmap fmtMessage logTextStdout
+     in run la [Message D callStack "message"])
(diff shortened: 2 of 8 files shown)


ChShersh:
@qnikst Thank you for these benchmarks! It's kinda strange that outputting `ByteString` to `stderr` is almost 10x times slower than outputting to `stdout` :thinking: 

I think it's okay to keep these particular benchmarks in separate package. They might be huge, so no need to move into existing package. I'm okay to not use `criterion` or any other benchmark library here.

co-log-benchmark/co-log-benchmark.cabal
```diff
@@ -0,0 +1,32 @@
+name:                co-log-benchmark
```
ChShersh: Can you please update `cabal.project` file and add name of this package there?
And also update `.travis.yml` file for `stack` section to build this package on CI as well

co-log-benchmark/Setup.hs
```diff
@@ -0,0 +1,2 @@
+import Distribution.Simple
+main = defaultMain
```
ChShersh: I think that `Setup.hs` is not needed for `cabal`, it works without this file pretty fine

co-log-benchmark/Main.hs
```diff
@@ -0,0 +1,103 @@
...
+  [] -> do
+     putStrLn "Dump 10k messages (in a forked process):"
+     for_ benchs $ \(name, _) -> do
+       t <- timeProcess name
+       printf "%-50s %s\n" name (show t)
+  (name:_) -> case name `lookup` benchs of
+     Nothing -> pure ()
+     Just f  -> f
+
+-- | Measure the running time of the process.
```
ChShersh: Thanks again for the documentation!

co-log-benchmark/LICENSE
```diff
@@ -0,0 +1,20 @@
+Copyright (c) 2018 Alexander Vershilov
```
ChShersh: I wonder, how should I update the LICENSE if I want to do something with the benchmarks (like, refactor or add new ones)? :thinking:

qnikst: oh.. that was autogenerated.

co-log-benchmark/Setup.hs
qnikst: that require you to have cabal-install. But with Setup.hs, you may install library without `cabal-install` that what most of the distributions do.

qnikst:
Seems `putBSLn` from relude is hitting some lucky pass.

co-log-benchmark/co-log-benchmark.cabal
qnikst: Yep, will do

co-log-benchmark/Setup.hs
ChShersh: @qnikst Could you, please, elaborate on this? How is it possible to install Haskell library without `cabal-install` and why one might want to do this?

qnikst: you can always run:

```
ghc Setup.hs
./Setup configure
./Setup build
./Setup install
```
It will not install any dependency, but you don't need any additional tool to do that. Source distribution like `Gentoo`, `NixOS` do that, and other when you prepare a binary package. Because dependencies and package management is solved by the special tools.

stack.yaml
```diff
@@ -3,6 +3,7 @@ resolver: lts-12.9
 packages:
   - co-log
   - co-log-core
+  - co-log-benchmark
```
ChShersh: You also need `co-log-benchmark` to `.travis.yml` here:

https://github.com/kowainik/co-log/blob/76964369c38152b2ad10bd3c452e0e7a0a90dfb9/.travis.yml#L79-L80

co-log-benchmark/co-log-benchmark.cabal
```diff
@@ -0,0 +1,32 @@
...
+
+source-repository    head
+  type:              git
+  location:          https://github.com/kowainik/co-log.git
+
+executable co-log-features-bench
+  main-is:             Main.hs
+  -- other-modules:
+  -- other-extensions:
+  build-depends:       base >=4.11 && <4.12,
```
ChShersh: This is needed to fix CI for different Cabal versions

```suggestion
  build-depends:       base >=4.10 && <4.13,
```

co-log-benchmark/Main.hs
```diff
@@ -0,0 +1,103 @@
...
+import System.Posix.Process
+import qualified Data.Text.Encoding
+import GHC.Stack
+import Text.Printf
+
+
+-- | List of benchmarks.
+benchs :: [(String, IO ())]
+benchs =
+  [("baseline", run (LogAction (\_ -> pure ())) ["message"::String])
```
ChShersh: `mempty` should work

```suggestion
  [("baseline", run mempty ["message"::String])
```

co-log-benchmark/co-log-benchmark.cabal
```diff
@@ -23,10 +22,11 @@ executable co-log-features-bench
   main-is:             Main.hs
   -- other-modules:
   -- other-extensions:
-  build-depends:       base >=4.10 && <4.13,
+  build-depends:       base-noprelude >=4.10 && <4.13,
```
ChShersh: Heh. `Prelude` in the `co-log` package is now in `other-modules` so `base-noprelude` trick won't work. Anyway, you didn't used `relude` before in `co-log-benchmarks` so I don't expect that addition of `relude` is required... Need to see CI failure to check what's wrong

co-log-benchmark/Main.hs
```diff
@@ -0,0 +1,103 @@
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings #-}
+module Main
+  ( main
+  ) where
+
+import Data.Time.Clock
```
ChShersh:
```suggestion
import Data.Semigroup ((<>))
import Data.Time.Clock
```

ChShersh: I hope that should fix CI error. But the last commit with `base-noprelude` should be reverted. Alternative, you can add `Prelude` module to the `co-log-benchmarks` packages

co-log-benchmark/co-log-benchmark.cabal
qnikst: I need to either use Prelude trick:
```
import Data.Semigroup
import Prelude
```

so semigroup will be imported in all GHC versions, or use `relude`.

Issue #63, opened 2018-10-13
Add `logShowStdout` to Colog.Core.IO


cronokirby:
This came up in #61 while resolving #58:

In the documentation examples it was often needed to log a value implementing `Show`, e.g. Int, (), etc.
I think this would a useful function to add to Colog.Core.IO.
`logShowStdout` is just a first idea of a name; it matches `logStringStdout`, but maybe we want `logPrint`
or something else instead.

ChShersh:
@cronokirby I'm actually okay with `logPrint` because by default `print` outputs to `stdout` so `logPrint` is short and self-explanatory name. If desired, we can add `logPrintStderr` though it kinda against convention... I propose to gather more feedback under this issue before implementing.

PR #66, opened 2018-10-21
[#63] Add `logPrint` and `logPrintStderr` to Colog.Core.IO.

gahag:
Resovles  #63:

Added the suggested `logPrint` and `logPrintStderr`.
Updated the benchmarks to test `logPrint`.
Updated example to use the new functions.


gahag:
Done.

PR #73, opened 2018-11-12
Upgrade to relude 0.4.0

### LINKED ISSUE #68: Bump up to `relude-0.4.0`
Checklist:

1. Update bounds in `.cabal` file.
2. Update in `stack.yaml`
3. Update HLint rules.
4. Add check for HLint rules to the Travis CI and fix all HLint warnings and hints.

cronokirby:
Resolves #68 

Upgrades the project to relude 0.4.0, and starts using relude's hlint rules.

Updates travis to download and run hlint with the rules provided [here](https://github.com/kowainik/relude/blob/master/.hlint.yaml), and fixes the various hlint issues around the project.

2 notable workarounds were needed to comply with lint in [Message.hs](https://github.com/kowainik/co-log/blob/master/co-log/src/Colog/Message.hs):
- `MessageField` couldn't be a GADT record to be parsed correctly by hlint
- `SrcLoc` had to have its record fields imported manually, because the import of `SrcLoc(..)` was
    causing hlint to think that other functions reexported from `GHC.Stack` by relude were being used
    from `GHC.Stack` itself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,7 +79,7 @@ script:
       stack build co-log-core --test --bench --no-run-benchmarks --no-terminal
       stack build co-log      --test --bench --no-run-benchmarks --no-terminal
     fi
-  - curl -sSL https://raw.github.com/ndmitchell/neil/master/misc/travis.sh | sh -s -- hlint -h .hlint-relude.yaml src/ test/    
+  - curl -sSL https://raw.github.com/ndmitchell/neil/master/misc/travis.sh | sh -s -- hlint -h .hlint-relude.yaml co-log/src/ co-log-core/src/    
 
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,7 +79,8 @@ script:
       stack build co-log-core --test --bench --no-run-benchmarks --no-terminal
       stack build co-log      --test --bench --no-run-benchmarks --no-terminal
     fi
-
+  - curl https://raw.githubusercontent.com/kowainik/relude/988bbdd3a09df1159917012933780523644880e5/.hlint.yaml -o .hlint-relude.yaml
+  - curl -sSL https://raw.github.com/ndmitchell/neil/master/misc/travis.sh | sh -s -- hlint -h .hlint-relude.yaml src/ test/    
 
 notifications:
   email: false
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -79,7 +79,8 @@ script:
       stack build co-log-core --test --bench --no-run-benchmarks --no-terminal
       stack build co-log      --test --bench --no-run-benchmarks --no-terminal
     fi
-
+  - curl https://raw.githubusercontent.com/kowainik/relude/988bbdd3a09df1159917012933780523644880e5/.hlint.yaml -o .hlint-relude.yaml
+  - curl -sSL https://raw.github.com/ndmitchell/neil/master/misc/travis.sh | sh -s -- hlint -h .hlint-relude.yaml src/ test/
```
vrom911: I guess it checks wrong locations, shouldn't it be `co-log` package's modules only?

cronokirby: Oh I guess we should keep it in sync with the hlint file we have here

cronokirby:
I guess we need to use -XCPP for GHC 8.6.1, and not import >$< from Colog

cronokirby:
The workaround I've settled on is to hide `>$` and `>$<` from Relude, which exports them for GHC 8.6.1.
There are a few places in colog that don't play well with the exports from `Data.Functor.Contravariant`, and which should look into those in another issue

.travis.yml
```diff
@@ -79,7 +79,7 @@ script:
       stack build co-log-core --test --bench --no-run-benchmarks --no-terminal
       stack build co-log      --test --bench --no-run-benchmarks --no-terminal
     fi
-
+  - curl -sSL https://raw.github.com/ndmitchell/neil/master/misc/travis.sh | sh -s -- hlint -h .hlint-relude.yaml co-log/src/
```
ChShersh: I think it's better to run `hlint` on `co-log-core` package as well but without `relude`-specific hints. The idea is the following:

1. Download `.hlint-relude.yaml` file via `curl`
2. Run raw `hlint` over `co-log/` package
3. Run `hlint` with `.hlint-relude.yaml` file over `co-log/` directory

So let's not add `.hlint-relude.yaml` file directly to this repository. Downloading it via `curl` during Travis is a way to go. Because we will also see the commit of `.hlint` file.

ChShersh:
Closing in favour of #74

PR #88, opened 2018-11-21
Added overLogAction to HasLog typeclass

### LINKED ISSUE #85: Move `overLogAction` into `HasLog` typeclass
And add `default` implementation to `setLogAction` function.

https://github.com/kowainik/co-log/blob/494238bb10dbf529140e9223416f9b200edc384a/co-log-core/src/Colog/Core/Class.hs#L26-L27

sphaso:
Resolves #85 

I'm not 100% certain that I did what you meant with:
> And add default implementation to setLogAction function.

### DIFF
--- a/co-log-core/CHANGELOG.md
+++ b/co-log-core/CHANGELOG.md
@@ -3,6 +3,11 @@
 `co-log-core` uses [PVP Versioning][1].
 The change log is available [on GitHub][2].
 
+## 0.1.2
+
+* [#85](https://github.com/kowainik/co-log/issues/85):
+  Move `overLogAction` to `HasLog` typeclass
+
 ## 0.1.1 — Nov 15, 2018
 
 * [#63](https://github.com/kowainik/co-log/issues/63):
--- a/co-log-core/src/Colog/Core/Class.hs
+++ b/co-log-core/src/Colog/Core/Class.hs
@@ -28,6 +28,6 @@ instance HasLog (LogAction m msg) msg m where
     {-# INLINE getLogAction #-}
     getLogAction = id
     {-# INLINE setLogAction #-}
-    setLogAction = flip const
+    setLogAction = const
     {-# INLINE overLogAction #-}
     overLogAction = id
--- a/co-log-core/src/Colog/Core/Class.hs
+++ b/co-log-core/src/Colog/Core/Class.hs
@@ -15,6 +15,7 @@ import Colog.Core.Action (LogAction)
 TODO: laws
 -}
 class HasLog env msg m where
+    {-# MINIMAL getLogAction, (setLogAction | overLogAction) #-}
     getLogAction :: env -> LogAction m msg
     setLogAction :: LogAction m msg -> env -> env
     setLogAction = overLogAction . const
@@ -22,8 +23,11 @@ class HasLog env msg m where
     overLogAction :: (LogAction m msg -> LogAction m msg) -> env -> env
     overLogAction f env = setLogAction (f $ getLogAction env) env
     {-# INLINE overLogAction #-}
-    {-# MINIMAL getLogAction, (setLogAction | overLogAction) #-}
 
 instance HasLog (LogAction m msg) msg m where
+    {-# INLINE getLogAction #-}
     getLogAction = id
+    {-# INLINE setLogAction #-}
+    setLogAction = flip const
+    {-# INLINE overLogAction #-}
     overLogAction = id
--- a/co-log-core/src/Colog/Core/Class.hs
+++ b/co-log-core/src/Colog/Core/Class.hs
@@ -17,9 +17,12 @@ TODO: laws
 class HasLog env msg m where
     getLogAction :: env -> LogAction m msg
     setLogAction :: LogAction m msg -> env -> env
+    setLogAction = seq
     overLogAction :: (LogAction m msg -> LogAction m msg) -> env -> env
+    overLogAction = seq
+    {-# MINIMAL getLogAction #-}
 
 instance HasLog (LogAction m msg) msg m where
     getLogAction = id
-    setLogAction = const
-    overLogAction over = over
+    setLogAction = overLogAction . const
+    overLogAction = id
--- a/co-log-core/src/Colog/Core/Class.hs
+++ b/co-log-core/src/Colog/Core/Class.hs
@@ -6,7 +6,6 @@
 
 module Colog.Core.Class
        ( HasLog (..)
-       , overLogAction
        ) where
 
 import Colog.Core.Action (LogAction)
@@ -18,11 +17,9 @@ TODO: laws
 class HasLog env msg m where
     getLogAction :: env -> LogAction m msg
     setLogAction :: LogAction m msg -> env -> env
+    overLogAction :: (LogAction m msg -> LogAction m msg) -> env -> env
 
 instance HasLog (LogAction m msg) msg m where
     getLogAction = id
     setLogAction = const
-
-overLogAction :: HasLog env msg m => (LogAction m msg -> LogAction m msg) -> env -> env
-overLogAction over env = setLogAction (over $ getLogAction env) env
-{-# INLINE overLogAction #-}
+    overLogAction over = over
--- a/co-log-core/src/Colog/Core/Class.hs
+++ b/co-log-core/src/Colog/Core/Class.hs
@@ -17,12 +17,13 @@ TODO: laws
 class HasLog env msg m where
     getLogAction :: env -> LogAction m msg
     setLogAction :: LogAction m msg -> env -> env
-    setLogAction = seq
+    setLogAction = overLogAction . const
+    {-# INLINE setLogAction #-}
     overLogAction :: (LogAction m msg -> LogAction m msg) -> env -> env
-    overLogAction = seq
-    {-# MINIMAL getLogAction #-}
+    overLogAction f env = setLogAction (f $ getLogAction env) env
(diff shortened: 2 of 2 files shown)


cronokirby:
@sphaso You can implement `setLogAction` via `overLogAction`, so it'd be good to have a default implementation in the typeclass that way, allowing users to specialise it if they have a more efficient solution.

I believe that's what was being requested in the issue

co-log-core/src/Colog/Core/Class.hs
```diff
@@ -18,11 +17,9 @@ TODO: laws
 class HasLog env msg m where
     getLogAction :: env -> LogAction m msg
     setLogAction :: LogAction m msg -> env -> env
+    overLogAction :: (LogAction m msg -> LogAction m msg) -> env -> env
```
ChShersh: Yes, you should add default implementations of `overLogAction` and `setLogAction`

co-log-core/src/Colog/Core/Class.hs
```diff
@@ -18,11 +17,9 @@ TODO: laws
 class HasLog env msg m where
     getLogAction :: env -> LogAction m msg
```
ChShersh: Also it's a good idea to add `{-# MINIMAL #-}` pragma to this typeclass.

co-log-core/src/Colog/Core/Class.hs
```diff
@@ -18,11 +17,9 @@ TODO: laws
...
+    overLogAction :: (LogAction m msg -> LogAction m msg) -> env -> env
 
 instance HasLog (LogAction m msg) msg m where
     getLogAction = id
     setLogAction = const
-
-overLogAction :: HasLog env msg m => (LogAction m msg -> LogAction m msg) -> env -> env
-overLogAction over env = setLogAction (over $ getLogAction env) env
-{-# INLINE overLogAction #-}
+    overLogAction over = over
```
ChShersh: Looks like this is also `id`
```suggestion
    overLogAction = id
```

sphaso:
@cronokirby @ChShersh    
- added change to CHANGELOG
- added MINIMAL pragma
- refactored `overLogAction`
- added (very naive) default implementations for `setLogAction` and `overLogAction`

co-log-core/src/Colog/Core/Class.hs
```diff
@@ -18,11 +17,12 @@ TODO: laws
 class HasLog env msg m where
     getLogAction :: env -> LogAction m msg
     setLogAction :: LogAction m msg -> env -> env
+    setLogAction = seq
```
cronokirby: Why `seq` instead of `flip const`? Why are we forcing evaluation of the LogAction?

co-log-core/src/Colog/Core/Class.hs
cronokirby: Shouldn't we just define one in terms of the other?
Is there a sensible default implementation for *both* at once?

co-log-core/src/Colog/Core/Class.hs
```diff
@@ -18,11 +17,12 @@ TODO: laws
...
+    {-# MINIMAL getLogAction #-}
 
 instance HasLog (LogAction m msg) msg m where
     getLogAction = id
-    setLogAction = const
-
-overLogAction :: HasLog env msg m => (LogAction m msg -> LogAction m msg) -> env -> env
-overLogAction over env = setLogAction (over $ getLogAction env) env
-{-# INLINE overLogAction #-}
+    setLogAction = overLogAction . const
```
cronokirby: IMO the default implementation of setLogAction should be this in the typeclass, and then we don't even have to define it in this trivial case here

co-log-core/src/Colog/Core/Class.hs
sphaso: sorry, I'm not sure I'm following, aren't `seq` and `flip const` equal?
```
>> :t seq
seq :: a -> b -> b
>> :t flip const
flip const :: b -> c -> c
```

sphaso:
@ChShersh from the looks of it, it might make more sense to have a default implementation of `setLogAction` and only require the user to set `overLogAction`. I think @cronokirby has a point that they could be implemented in terms of each other as a default.

co-log-core/src/Colog/Core/Class.hs
ChShersh: By default I mean no `default` keyword, just like for `Eq` typeclass where `==` is implemented using `/=` and `/=` is implemented using `==`

co-log-core/CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 `co-log-core` uses [PVP Versioning][1].
 The change log is available [on GitHub][2].
 
+## 0.1.2
+
+* [#85](https://github.com/kowainik/co-log/issues/85):
+  Move `overLogAction` to `HasLog` typeclass
```
ChShersh: I wonder whether it's a breaking change and should go to `0.2.0` or not 🤔

co-log-core/src/Colog/Core/Class.hs
ChShersh: `setLogAction` here should be implemented using `overLogAction`.

Also, don't forget to add `{-# INLINE #-}` pragmas to the default implementations 🙂

co-log-core/src/Colog/Core/Class.hs
```diff
@@ -18,11 +17,12 @@ TODO: laws
 class HasLog env msg m where
     getLogAction :: env -> LogAction m msg
     setLogAction :: LogAction m msg -> env -> env
+    setLogAction = seq
+    overLogAction :: (LogAction m msg -> LogAction m msg) -> env -> env
+    overLogAction = seq
```
ChShersh: Also, not `seq`, you should use `getLogAction` and `setLogAction` to implement `overLogAction`

co-log-core/src/Colog/Core/Class.hs
```diff
@@ -18,11 +17,12 @@ TODO: laws
...
+    {-# MINIMAL getLogAction #-}
```
ChShersh: Minimal complete set actually should be:

```haskell
{-# MINIMAL getLogAction, (setLogAction | overLogAction) #-}
```

sphaso:
@cronokirby @ChShersh hopefully I didn't make any egregious mistake this time :) thanks for your review as always!

co-log-core/src/Colog/Core/Class.hs
```diff
@@ -18,11 +17,13 @@ TODO: laws
...
+    {-# MINIMAL getLogAction, (setLogAction | overLogAction) #-}
 
 instance HasLog (LogAction m msg) msg m where
     getLogAction = id
-    setLogAction = const
-
-overLogAction :: HasLog env msg m => (LogAction m msg -> LogAction m msg) -> env -> env
-overLogAction over env = setLogAction (over $ getLogAction env) env
-{-# INLINE overLogAction #-}
+    overLogAction = id
```
ChShersh: Let's also implement `setLogAction` function. If you don't implement methods with default definitions from typeclass, then the default implementation of those methods is used. If you implement methods manually, you have better control over the implementation and thus you can have better performance. Also, let's add `{-# INLINE #-}` pragmas to all three functions in this instance to have even more guarantees for performance.

co-log-core/src/Colog/Core/Class.hs
```diff
@@ -18,11 +17,13 @@ TODO: laws
 class HasLog env msg m where
     getLogAction :: env -> LogAction m msg
     setLogAction :: LogAction m msg -> env -> env
+    setLogAction = overLogAction . const
+    {-# INLINE setLogAction #-}
+    overLogAction :: (LogAction m msg -> LogAction m msg) -> env -> env
+    overLogAction f env = setLogAction (f $ getLogAction env) env
+    {-# INLINE overLogAction #-}
+    {-# MINIMAL getLogAction, (setLogAction | overLogAction) #-}
```
ChShersh: Can you, please, run haddock command for `co-log-core` and post screenshot of how it looks like? I hope that syntax is correct and there should be no problems with rendering it... Also, it's usually a good practice to put `{-# MINIMAL #-}` pragma before all methods inside typeclass definition.

sphaso:
@ChShersh 
- added INLINE everywhere
- moved MINIMAL on top
- implemented `setLogAction` in the instance

This is the [Haddock output in HTML](https://imgur.com/a/gOiPGkI)

co-log-core/src/Colog/Core/Class.hs
```diff
@@ -16,13 +15,19 @@ import Colog.Core.Action (LogAction)
...
 instance HasLog (LogAction m msg) msg m where
+    {-# INLINE getLogAction #-}
     getLogAction = id
-    setLogAction = const
-
-overLogAction :: HasLog env msg m => (LogAction m msg -> LogAction m msg) -> env -> env
-overLogAction over env = setLogAction (over $ getLogAction env) env
-{-# INLINE overLogAction #-}
+    {-# INLINE setLogAction #-}
+    setLogAction = flip const
```
ChShersh: Looks like a minor error in the implementation
```suggestion
    setLogAction = const
```

ChShersh:
@sphaso Let's keep it minor version bump up. I tried to think about this more and looks like it's not a breaking change. `overLogAction` has default definition, so existing instances won't break. Import `Colog (HasLog (..), overLogAction)` doesn't result into error, so it's still a valid Haskell code.

Issue #76, opened 2018-11-14
Add benchmarks result to the readme


ChShersh:
Ideally I would like to see pure function that takes list of pairs -- benchmark name and result and generates markdown table which we can then insert to the `README` easily in case we want to update benchmark results.

C37H40O9:
Something like this?

```haskell
import Data.List (intercalate)
import Data.Monoid ((<>))

genTable pairs = intercalate "\n" $ header <> rows
    where
      header = intercalate " | " <$> [["Benchmark", "Result"], ["---","---:"]]
      rows = fmap (\(b,r) -> b <> " | " <> show r) pairs
```
For `[("B1",1),("B2",12),("B3",123)]` it produce
```markdown
Benchmark | Result
--- | ---:
B1 | 1
B2 | 12
B3 | 123
```

Benchmark | Result
--- | ---:
B1 | 1
B2 | 12
B3 | 123

C37H40O9:
Version with aligning:
```haskell
import Data.List (intercalate)
import Data.Semigroup((<>),Max(..))
import Data.Bifunctor (bimap, second)
import Data.Coerce (coerce)

genTable pairs = intercalate "\n" rows
  where
    pairs' = ("Benchmark", "Result") : ("---","---:") : fmap (second show) pairs
    f = (Max).length
    (fMax,sMax) = coerce $ foldMap (bimap f f) $ fmap (second show) pairs'
    rows = fmap (\(b,r) -> b <> replicate (fMax - (length b)) ' ' <> " | " 
                             <> replicate (sMax - (length r)) ' ' <> r) pairs'
```
produce
```
Benchmark             |   Result
---                   |     ---:
VeryLongBenchmarkName |        1
ShortBench            |       12
MediumBenchmark       |      123

```
Benchmark             |   Result
---                   |     ---:
VeryLongBenchmarkName |        1
ShortBench            |       12
MediumBenchmark       |      123

ChShersh:
@C37H40O9 this looks nice! You're welcome to open PR if you want 🙂

C37H40O9:
Cool! My (first) PR waiting for review :-)

ChShersh:
The function for formatting is implemented in #90. I will refactor code of benchmarks a little bit (by adding description to benchmarks, etc.) and then I will insert the results into README.

ChShersh:
Closed by #92

PR #90, opened 2018-12-04
[#76] Added function for generating markdown table with benchmarks

C37H40O9:


co-log-benchmark/Main.hs
```diff
@@ -106,3 +109,13 @@ timeProcess n = do
       runProcess_ cfg
   t' <- getCurrentTime
   pure $ t' `diffUTCTime` t
+
+-- |  Function that takes list of pairs - benchmark name
+-- and result and generates markdown table.
+genTable pairs = intercalate "\n" rows
```
ChShersh: It would be really nice for this function to have a type signature :slightly_smiling_face:

co-log-benchmark/Main.hs
```diff
@@ -106,3 +109,13 @@ timeProcess n = do
       runProcess_ cfg
   t' <- getCurrentTime
   pure $ t' `diffUTCTime` t
+
+-- |  Function that takes list of pairs - benchmark name
+-- and result and generates markdown table.
```
ChShersh: It would be much better if `main` uses this `genTable` function to not just print the results of current benchmarks to the terminal, but also collect all results and only then output them as markdown table. So later everyone can just run the benchmarks and copy-paste the table to README.

co-log-benchmark/Main.hs
```diff
@@ -106,3 +109,13 @@ timeProcess n = do
...
+genTable pairs = intercalate "\n" rows
+  where
+    pairs' = ("Benchmark", "Result") : ("---","---:") : fmap (second show) pairs
+    f = (Max).length
+    (fMax,sMax) = coerce $ foldMap (bimap f f)  pairs'
+    rows = fmap (\(b,r) -> b <> replicate (fMax - (length b)) ' ' <> " | "
```
cronokirby: I'd probably pull out another subfunction for the `replicate` part, in order to make rows a bit more clear
type signatures for these subfunctions wouldn't hurt

C37H40O9:
Huh, there was a little problem due to the fact that building project on Nix-OS turned out to be a challenge for me.
Now co-log-features-bench executable run benchmarks and print table with results

Benchmark                                      |       Result
---                                            |         ---:
baseline                                       | 0.001729273s
stdout<print                                   | 0.004978595s
stdout<string                                  | 0.009066414s
stdout<text                                    |  0.00922734s
stdout<bytestring                              |  0.00500579s
stderr<bytestring                              | 0.009105494s
{stdout<>stderr}<bytestring                    | 0.009085795s
stdout<format<message                          | 0.009109447s
stdout<format<message{with callstack}          | 0.009111841s
stdout<format<message{with callstack:5}        | 0.009094889s
stdout<format<message{with callstack:50}       | 0.009040144s
stdout<bytestring<format<message               | 0.004857813s
stdout<bytestring<rich-message<upgrade<message | 0.234826368s

C37H40O9:
Strange that on the previous commit stack build passed while the others failed.

PR #94, opened 2019-01-13
[#77] Use chronos time formatter

### LINKED ISSUE #77: Consider using faster time formatting
There're benchmarks in `chronos` library regarding time pretty printing:

* https://github.com/andrewthad/chronos/pull/24

`katip` logging library also has fast implementation of time formatting:

* https://github.com/Soostone/katip/blob/e04f087f5058f815b37b6a9b2ceafb366e45eec2/katip/src/Katip/Format/Time.hs

lucazulian:
Resolves #77


co-log/co-log.cabal
```diff
@@ -52,6 +52,7 @@ library
                      , stm >= 2.4 && < 2.6
                      , text ^>= 1.2.3
                      , time >= 1.8 && < 1.10
+                     , chronos >= 1.0.4
```
ChShersh: It's always better to add upper bounds as well for better compatibility.

```suggestion
                     , chronos ^>= 1.0.4
```

co-log/co-log.cabal
```diff
@@ -52,6 +52,7 @@ library
                      , stm >= 2.4 && < 2.6
                      , text ^>= 1.2.3
                      , time >= 1.8 && < 1.10
```
ChShersh: `chronos` should replace `time` dependency, so I guess this library can be safely removed from `build-depends`

co-log/src/Colog/Message.hs
```diff
@@ -139,7 +139,7 @@ types. The type family is open so you can add new instances.
 -}
 type family FieldType (fieldName :: Symbol) :: Type
 type instance FieldType "threadId" = ThreadId
-type instance FieldType "utcTime"  = UTCTime
+type instance FieldType "utcTime"  = C.Time
```
ChShersh: According to the documentation for `Time` data type from the `chronos` library, this is POSIX time, not UTC time. 

* https://hackage.haskell.org/package/chronos-1.0.4/docs/Chronos-Types.html#t:Time

So I guess this type family instance should be also changed, from `"utcTime"` to something like `"posixTime"` or `"chronosTime"`.

co-log/src/Colog/Message.hs
```diff
@@ -227,19 +227,19 @@ fmtRichMessageDefault RichMessage{..} = do
...
      <> thread
      <> messageText
 
-    showTime :: UTCTime -> Text
-    showTime t = square $ T.pack $
-          formatTime defaultTimeLocale "%H:%M:%S." t
-       ++ take 3 (formatTime defaultTimeLocale "%q" t)
-       ++ formatTime defaultTimeLocale " %e %b %Y %Z" t
+    showTime :: C.Time -> Text
+    showTime t = square $ T.pack $ show $ C.builder_DmyHMS timePrecision datetimeFormat (C.timeToDatetime t)
```
ChShersh: Wow, I didn't expect that writing formatter with `chronos` would be that simple! Nice work :+1:

co-log/src/Colog/Message.hs
```diff
@@ -227,19 +227,19 @@ fmtRichMessageDefault RichMessage{..} = do
...
+        where
+          timePrecision = C.SubsecondPrecisionFixed 3
+          datetimeFormat = C.DatetimeFormat (Just ' ') (Just ' ') (Just ':')
```
ChShersh: According to our style guide, 3 lines above shifted 2 spaces left :slightly_smiling_face:

stack.yaml
```diff
@@ -4,3 +4,7 @@ packages:
   - co-log
   - co-log-core
   - co-log-benchmark
+
+extra-deps:
+  - chronos-1.0.4
+  - torsor-0.1
```
ChShersh: Addition of these dependencies have one problem for `co-log`: it's not possible to add `co-log` to Stackage while these two packages are not there :disappointed: 

I've already opened issue for `torsor`:

* https://github.com/andrewthad/torsor/issues/1

I guess the same should be done for `chronos` as well.

co-log/src/Colog/Message.hs
ChShersh: Though, using `show` for Builder looks like a performance problem :thinking: I think it's better to use `toLazyText` and `fromLazy` here instead of `show` and `T.pack`:

* https://hackage.haskell.org/package/text-1.2.3.0/docs/Data-Text-Lazy-Builder.html#v:toLazyText

stack.yaml
lucazulian: I just opened the same issue for `chronos`
- https://github.com/andrewthad/chronos/issues/27

vrom911:
Thanks for the contribution, @lucazulian !

I'm curious how this affects benchmarks, did you had a chance to run it?

co-log/src/Colog/Message.hs
lucazulian: I think it's better to use `Data.Text.Lazy.toStrict $ Data.Text.Lazy.Builder.toLazyText`

ChShersh:
@vrom911 I've just launched benchmarks on my machine locally. The running time for the relevant case decreased from `174ms` to `54ms`. This is a huge win!

lucazulian:
@ChShersh @vrom911 what are these cases? How can I run these benchmarks?

ChShersh:
@lucazulian You can run benchmarks by executing command like:

```
cabal new-exec co-log-bench
```

Benchmarks are implemented in the `co-log-benchmarks` packages. The only case that uses time (and time formatting as a consequence) is the last one with the name `Message{Time,ThreadId} > format > stdout`.

stack.yaml
lucazulian: https://github.com/commercialhaskell/stackage/pull/4296

ChShersh: Nice! I see that it's already on Stackage. So you can bump up resolver version in STACK_YAML:

* https://www.stackage.org/nightly-2019-01-14

Also, since it's a breaking change, we need to bump up `co-log` version from `0.2.0` to `0.3.0` and add entry to changelog about this change.

ChShersh:
@cronokirby There're some plans to add Restyled to our projects (and run/check `stylish-haskell` automatically). See this issue:

* https://github.com/kowainik/org/issues/9

I hope I can find time to work on this :slightly_smiling_face:

lucazulian:
@ChShersh @cronokirby if you explain me how to format the project I can try to do by myself in this on in another PR

co-log/CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 `co-log` uses [PVP Versioning][1].
 The change log is available [on GitHub][2].
 
+## 0.3.0 — Jan 14, 2019
```
ChShersh: Let's put `unreleased` tag here to specify more explicitly that this version of the library is not released. Usually we tell the date only after we upload library to Hackage.

```suggestion
## Unreleased: 0.3.0
```

stack.yaml
```diff
@@ -3,4 +3,4 @@ resolver: lts-13.2
 packages:
```
ChShersh: In order to make it build with `stack` on CI you also need to update `resolver` here. Just replace `lts-13.2` with `nightly-2019-01-15`

ChShersh:
@lucazulian I'm using Spacemacs as my editor and I've configured it to apply `stylish-haskell` on save. I don't think it's convenient enough to do all this stuff manually... It's just an executable, so if you want to use you need to install it first (using cabal or stack) and then you can call from shell something like this:

```shell
stylish-haskell -i co-log/src/Colog/Message.hs
```

But probably you can configure `stylish-haskell` to be applied on save for your editor as well.

vrom911:
Thanks for this very useful improvement, @lucazulian, and very well done!

@ChShersh, one small thing, now we update the benchmarks in README. Can you put the new info as you already run it?

PR #105, opened 2019-05-03
[#101] Add logActionL lens, impove docs a bit

### LINKED ISSUE #101: Add Lens' to the HasLog typeclass
The `HasLog` typeclass has all functions required for implementing custom lens. Since lenses are reusable, because they are just type alias, I propose to add the following field to the `HasLog` typeclass with the default definition:

```haskell
logActionL :: Lens' env (LogAction m msg)
logActionL = ...
```

So we need to introduce our own `Lens'` type alias (and reexport it) and add `logActionL` lens to the `HasLog` typeclass with the default definition (so people can implement their own definition if they want.

With this approach we won't depend on any `*lens*` package but still will provide lens-like interface using ideas from the following blog post:

* https://vrom911.github.io/blog/write-yourself-a-lens

vrom911:
Resolves #101

### DIFF
--- a/co-log-core/src/Colog/Core/Class.hs
+++ b/co-log-core/src/Colog/Core/Class.hs
@@ -21,6 +21,9 @@ module Colog.Core.Class
 import Colog.Core.Action (LogAction)
 
 
+-- to inline lens better
+{- HLINT ignore "Redundant lambda" -}
+
 {- | This types class contains simple pair of getter-setter and related
 functions.
 It also provides the useful lens 'logActionL' with the default implementation using type
--- a/co-log-core/src/Colog/Core/Class.hs
+++ b/co-log-core/src/Colog/Core/Class.hs
@@ -1,28 +1,48 @@
 {-# LANGUAGE FlexibleInstances     #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE Rank2Types            #-}
 
-{- | Provides type class for objects that has access to 'LogAction'.
+{- |
+Copyright:  (c) 2018-2019 Kowainik
+License:    MIT
+Maintainer: Kowainik <xrom.xkov@gmail.com>
+
+Provides type class for objects that has access to 'LogAction'.
 -}
 
 module Colog.Core.Class
        ( HasLog (..)
+
+         -- * Lens
+         -- $lens
+       , Lens'
        ) where
 
 import Colog.Core.Action (LogAction)
 
-{- | This types class contains simple pair of getter-setter.
+
+{- | This types class contains simple pair of getter-setter and related
+functions.
+It also provides the useful lens 'logActionL' with the default implementation using type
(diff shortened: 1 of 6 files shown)


co-log-core/src/Colog/Core/Class.hs
```diff
@@ -1,33 +1,72 @@
...
     overLogAction :: (LogAction m msg -> LogAction m msg) -> env -> env
     overLogAction f env = setLogAction (f $ getLogAction env) env
     {-# INLINE overLogAction #-}
 
+    logActionL :: Lens' env (LogAction m msg)
+    logActionL = lens getLogAction (flip setLogAction)
+    {-# INLINE logActionL #-}
+
 instance HasLog (LogAction m msg) msg m where
     {-# INLINE getLogAction #-}
```
chshersh: Could I, please, ask you to move `{-# INLINE #-}` pragmas in this instance after definitions? Just to make it more consistent with out style guide 🙂 I'm not sure why it was done like this...

co-log-core/src/Colog/Core/Class.hs
```diff
@@ -1,33 +1,72 @@
...
     getLogAction = id
     {-# INLINE setLogAction #-}
     setLogAction = const
     {-# INLINE overLogAction #-}
     overLogAction = id
```
chshersh: While we are on this, could you please add implementation of `logActionL` to this instance? Just to use more efficient implementation instead of the default one.

co-log-core/src/Colog/Core/Class.hs
```diff
@@ -1,33 +1,72 @@
...
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE Rank2Types            #-}
 
-{- | Provides type class for objects that has access to 'LogAction'.
+{- |
+Copyright:  (c) 2018-2019 Kowainik
+License:    MIT
+Maintainer: Kowainik <xrom.xkov@gmail.com>
+
+Provides type class for objects that has access to 'LogAction'.
```
chshersh: Not sure what I was thinking before 🤔 
```suggestion
Provides type class for values that has access to 'LogAction'.
```

co-log-core/src/Colog/Core/Severity.hs
```diff
@@ -1,13 +1,31 @@
...
 
-{- | This module introduces 'Severity' data type for expressing how severe the
-message is. Also, it contains useful functions for work with 'Severity'.
+{- |
+Copyright:  (c) 2018-2019 Kowainik
+License:    MIT
+Maintainer: Kowainik <xrom.xkov@gmail.com>
 
+This module introduces 'Severity' data type for expressing how severe the
+message is. Also, it contains useful functions and patterns for work with 'Severity'.
```
chshersh: Documentation for this module is extremely nice!

PR #109, opened 2019-05-03
[#99] Add more comonadic combinators

### LINKED ISSUE #99: Add `duplicate` combinator to `co-log-core`
From the `Comonad` typeclass.

chshersh:
Resolves #99

### DIFF
--- a/co-log-core/src/Colog/Core/Action.hs
+++ b/co-log-core/src/Colog/Core/Action.hs
@@ -471,14 +471,14 @@ in duplicate logger <& ([3, 4], [42, 10])
 
 __Implementation note:__
 
-True and fair translation of the @duplication@ function from the 'Comonad'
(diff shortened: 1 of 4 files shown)


co-log-core/src/Colog/Core/Action.hs
```diff
@@ -442,3 +455,66 @@ infixr 1 <<=
...
+msg -> msg -> m ()
+@
+
+In order to capture this behavior, 'duplicate' should have the following type:
+
+@
+duplicate :: Semigroup msg => LogAction m msg -> LogAction (Compose ((->) msg) m) msg
+@
+
+However, it's quite awkward to work with such type. It's as known fact that the
```
vrom911:
```suggestion
However, it's quite awkward to work with such type. It's a known fact that the
```

Issue #84, opened 2018-11-16
Create series of examples with tutorials


chshersh:
> Currently all tutorials are here: https://github.com/kowainik/co-log/tree/master/co-log/example

Proposed by @vrom911. Basically, have `examples` directory with different complete usage examples of `co-log` library. I propose to have the following examples:

- [x] Basic example where you pass `LogAction IO String` as an argument and feed messages to it with `<&` operator
- [ ] Same example but with `LoggerT` monad and simple message
- [ ] Same example but with `LoggerT` monad and rich message
- [ ] Basic example with `LoggerT` but not printing `threadId`
- [ ] Like with `LoggerT` monad but using `withLog` function to change logger action locally
- [ ] Using `PureLogger`
- [x] Using custom monad with `LogAction` inside `Env`
- [ ] Logger combinators
- [ ] Logger rotation (#133)

Should be good for now :+1: Maybe there will be some code duplication between examples but I think that it's better to have all code in single module for the example to be able to see full code.

Not to say that each example should contain pretty good documentation :smirk: 

Things to do after finishing:

- [ ] Remove `play-co-log` executable
- [x] Rename `example` directory to `tutorials`
- [ ] Change README for `co-log`

pangwa:
+1, is there any example on terminal logging and log rotation?

chshersh:
@pangwa I agree that having an example of logging rotation would be good. I've added it to the list (and I see that you've also opened a separate issue and PR, will review in a moment). Regarding logging to terminal: currently, all examples are using logging to the terminal.

PR #115, opened 2019-05-05
[#84] Write intro tutorial

chshersh:


co-log/example/1-intro/Intro.md
```diff
@@ -0,0 +1,82 @@
...
+`co-log` is based on the following data type:
+
+```idris
+newtype LogAction m msg = LogAction
+    { unLogAction :: msg -> m ()
+    }
+```
+
+Logging action is a function from some message of user-defined type `msg` that
+perform all logic inside some monad `m`. In `co-log` logging represented as a
```
vrom911:
```suggestion
perform all logic inside some monad `m`. In the `co-log` library logging is represented as a
```

co-log/example/1-intro/Intro.md
```diff
@@ -0,0 +1,82 @@
...
+value. With approach you can modify the way you do logging by simply performing
```
vrom911:
```suggestion
value. With such approach, you can modify the way you do logging by simply performing some
```

co-log/example/1-intro/Intro.md
```diff
@@ -0,0 +1,82 @@
...
+transformations with the value you have.
+
+Let's first look at very basic example of using `LogAction`:
```
vrom911:
```suggestion
Let's first look at a very basic example of using `LogAction`:
```

co-log/example/1-intro/Intro.md
```diff
@@ -0,0 +1,82 @@
...
+
+```haskell
+example1 :: LogAction IO String -> IO ()
+example1 logger = do
+    unLogAction logger "Example 1: First message"
+    unLogAction logger "Example 1: Second message"
+```
+
+If you want to do logging, you need to pass `LogAction` as an argument to your
```
vrom911: This is a too strict statement. Probably better to say that passing the arguments id only **one of** the options, and it is the simplest one.

co-log/example/1-intro/Intro.md
```diff
@@ -0,0 +1,82 @@
...
+function. In this example we used `LogAction` that takes `String`s as messages
```
vrom911:
```suggestion
function. In this example, we are using `LogAction` that takes `String`s as messages
```

co-log/example/1-intro/Intro.md
```diff
@@ -0,0 +1,82 @@
...
+and performs logging inside `IO` monad.
+
+For convenience, library defines useful operator `<&` that makes logging easier:
```
vrom911: Probably not logging, but it makes code easier?

co-log/example/1-intro/Intro.md
```diff
@@ -0,0 +1,82 @@
...
+
+```haskell
+example2 :: LogAction IO String -> IO ()
+example2 logger = do
+    logger <& "Example 2: First message"
+    logger <& "Example 2: Second message"
+```
+
+In order to do some logging, we need to pass some `logger` to our functions.
```
vrom911: Before you already wrote that you do logging, but here you say it like this is the function that would make logging, not the previous ones, probably could be rephrased, or some previous sentences should be modified? I'm a bit lost what is logging and what is not.

co-log/example/1-intro/Intro.md
```diff
@@ -0,0 +1,82 @@
...
+Here we are going to use the following `LogAction`:
+
+```idris
+logStringStdout :: LogAction IO String
+```
+
+This action just uses `putStrLn` underhood.
```
vrom911:
```suggestion
This action uses `putStrLn` underhood.
```

vrom911: Also probably mention `stdout`.

co-log/example/1-intro/Intro.md
```diff
@@ -0,0 +1,82 @@
...
+
+```haskell
+main :: IO ()
+main = do
+    let logger = logStringStdout
+    example1 logger
+    example2 logger
+```
```
vrom911: Probably comments before this piece of code about what is going on (like putting all together)?

co-log/example/1-intro/Intro.md
```diff
@@ -0,0 +1,109 @@
...
+    { unLogAction :: msg -> m ()
+    }
+```
+
+Logging action is a function from some message of user-defined type `msg` that
+performs all logic inside some monad `m`. In the `co-log` library **logger** is represented as a
+value. With such approach, you can modify the way you do logging by simply performing some
+transformations with the value you have.
+
+Let's first look at a very basic example of using just `putStrLn` for logging:
```
vrom911:
```suggestion
Let's first look at a very basic example of simply using `putStrLn` for logging:
```

co-log/example/1-intro/Intro.md
```diff
@@ -0,0 +1,109 @@
...
+```haskell
+example0 :: IO ()
+example0 = do
+    putStrLn "Example 0: First message"
+    putStrLn "Example 0: Second message"
+```
+
+Using `putStrLn` for logging is a very simple and basic approach for logging.
+When your application becomes bigger and more complex, you might want to bring
+some logging library into it.
```
vrom911: probably add reasons why would you want that, like, to let the library handle more difficult cases or anything like that?

co-log/example/1-intro/Intro.md
```diff
@@ -0,0 +1,109 @@
...
+
+Now let's look at how you can `LogAction` instead of `putStrLn` to achieve the
```
vrom911:
```suggestion
Now let's look at how you can use `LogAction` instead of `putStrLn` to achieve the
```

co-log/example/1-intro/Intro.md
```diff
@@ -0,0 +1,109 @@
...
+same goal. With `co-log` you need to have value of type `LogAction` that defines
```
vrom911:
```suggestion
same goal. With `co-log` you need to have a value of type `LogAction` that defines
```

co-log/example/1-intro/Intro.md
```diff
@@ -0,0 +1,109 @@
...
+how you are going to do logging. So you configure your logging separately and
```
vrom911:
```suggestion
how you are going to do logging. So you configure your logging settings separately and
```

co-log/example/1-intro/Intro.md
```diff
@@ -0,0 +1,109 @@
...
+details:
+
+```haskell
+example1 :: LogAction IO String -> IO ()
+example1 logger = do
+    unLogAction logger "Example 1: First message"
+    unLogAction logger "Example 1: Second message"
+```
+
+If you want to do logging with `co-log`, then of the options (and the simplest one)
```
vrom911:
```suggestion
If you want to do logging with `co-log`, then one of the options (and the simplest one)
```

co-log/example/1-intro/Intro.md
```diff
@@ -0,0 +1,109 @@
...
+
+Using `putStrLn` for logging is a very simple and basic approach for logging.
+When your application becomes bigger and more complex, you might want to bring
+some logging library into it.
+
+Now let's look at how you can `LogAction` instead of `putStrLn` to achieve the
+same goal. With `co-log` you need to have value of type `LogAction` that defines
+how you are going to do logging. So you configure your logging separately and
+then pass and use this `LogAction` value. See the following example for more
+details:
```
vrom911: Probably, it's better to point how this function is almost the same as the example0 and say something about how `logger` could look like, or say that you will cover it later? I guess it could help understand all together better, or how do you think?

chshersh: I think it's a good idea to say how the logging can look like, so readers can have better understanding. Thanks for the suggestion!

PR #117, opened 2019-05-05
[#84] Write tutorial about using co-log with custom monad

chshersh:


co-log/example/2-custom/Custom.md
```diff
@@ -0,0 +1,135 @@
...
+              richMessageAction)
+import Control.Monad.IO.Class (MonadIO)
+import Control.Monad.Reader (MonadReader, ReaderT (..))
+```
+
+## Application environment
+
+If you have complex Haskell application, then most likely you also have
+non-trivial settings that configure your application environment. The
+environment may store various parameters important for work your application.
```
vrom911:
```suggestion
environment may store various parameters important for the work of your application.
```

co-log/example/2-custom/Custom.md
```diff
@@ -0,0 +1,135 @@
...
+
+```haskell
+simpleEnv :: Env App
+simpleEnv = Env
+    { envServerPort = 8081
+    , envLogAction  = richMessageAction
+    }
+```
+
+Then we need to define function that performs actions of type `App`:
```
vrom911:
```suggestion
Then we need to define a function that performs actions of type `App`:
```

co-log/example/2-custom/Custom.md
```diff
@@ -0,0 +1,135 @@
...
+and run our example.
+
+```haskell
+main :: IO ()
+main = runApp simpleEnv example
+```
+
+And the output will look like this:
+
+![Output example](https://user-images.githubusercontent.com/4276606/57191747-79579200-6f5b-11e9-92ed-e7b728a2b9a3.png)
```
vrom911: I didn't notice that the time formatting is changed! I think that this is the `chronos` library's formatting?

chshersh: Oops, yea, that might be the reason. Need to investigate formatting in this library...

PR #128, opened 2019-05-21
[#119] new action that logs a message without the severity information

### LINKED ISSUE #119: [RFC] Add new action that allows to print text messages with timestamp and thread id but without severity
Currently in order to print message with timestamp, you need to use `Message` data type which forces you to also specify severity. However, I may see that it might be more convenient sometimes to specify only textual payload and get the timestamp (and maybe thread id) automatically. This may help with moving from `putStrLn` to `co-log` easier.

The way to resolve this issue that I see is to change `RichMessage` data type. Currently it has the following shape:

https://github.com/kowainik/co-log/blob/1d700e04d73b544b871d42322e2f516f3b78c22a/co-log/src/Colog/Message.hs#L264-L268

I propose to perform the same thing that was done with the `Message` data type, specifically, change it to the following data type:

```haskell
data RichMsg (m :: Type -> Type) (msg :: Type) = RichMessage
    { richMsgMsg :: {-# UNPACK #-} !msg
    , richMsgMap :: {-# UNPACK #-} !(FieldMap m)
    } deriving (Functor)

type RichMessage m = RichMsg m Message
```

We need to change types of couple extra functions and implement some formatting in case of textual data (the last one is tricky). Otherwise should be straightforward.

sphaso:
Resolves #119 

- I refactored a bit trying to avoid duplicate code as per hlint suggestions
- I ran play-colog and I can in fact see the logs without the severity information

### DIFF
--- a/co-log/src/Colog/Message.hs
+++ b/co-log/src/Colog/Message.hs
@@ -294,7 +294,7 @@ __Examples:__
 See 'fmtMessage' if you don't need both time and thread id.
 -}
 fmtRichMessageDefault :: MonadIO m => RichMessage m -> m Text
-fmtRichMessageDefault msg = fmtRichMessageCustom msg formatRichMessage
+fmtRichMessageDefault msg = fmtRichMessageCustomDefault msg formatRichMessage
   where
     formatRichMessage :: Maybe ThreadId -> Maybe C.Time -> Message -> Text
     formatRichMessage (maybe "" showThreadId -> thread) (maybe "" showTime -> time) Msg{..} =
@@ -319,31 +319,31 @@ __Examples:__
 
 Practically, it formats a message as 'fmtRichMessageDefault' without the severity information.
 -}
-fmtSimpleRichMessageDefault :: MonadIO m => RichMessage m -> m Text
-fmtSimpleRichMessageDefault msg = fmtRichMessageCustom msg formatRichMessage
-  where
-    formatRichMessage :: Maybe ThreadId -> Maybe C.Time -> Message -> Text
-    formatRichMessage (maybe "" showThreadId -> thread) (maybe "" showTime -> time) Msg{..} =
+fmtSimpleRichMessageDefault :: MonadIO m => RichMsg m SimpleMsg -> m Text
+fmtSimpleRichMessageDefault msg = fmtRichMessageCustomDefault msg formatRichMessage
+   where
+    formatRichMessage :: Maybe ThreadId -> Maybe C.Time -> SimpleMsg -> Text
+    formatRichMessage (maybe "" showThreadId -> thread) (maybe "" showTime -> time) SimpleMsg{..} =
         time
-     <> showSourceLoc msgStack
+     <> showSourceLoc simpleMsgStack
      <> thread
-     <> msgText
+     <> simpleMsgText
 
-fmtRichMessageCustom :: MonadIO m => RichMessage m -> (Maybe ThreadId -> Maybe C.Time -> Message -> Text) -> m Text
-fmtRichMessageCustom RichMessage{..} formatter = do
+fmtRichMessageCustomDefault :: MonadIO m => RichMsg m msg -> (Maybe ThreadId -> Maybe C.Time -> msg -> Text) -> m Text
+fmtRichMessageCustomDefault RichMessage{..} formatter = do
     maybeThreadId  <- extractField $ TM.lookup @"threadId"  richMsgMap
     maybePosixTime <- extractField $ TM.lookup @"posixTime" richMsgMap
     pure $ formatter maybeThreadId maybePosixTime richMsgMsg
 
 showTime :: C.Time -> Text
 showTime t =
-            square
-          $ toStrict
-          $ TB.toLazyText
-          $ C.builder_DmyHMS timePrecision datetimeFormat (C.timeToDatetime t)
-        where
-            timePrecision = C.SubsecondPrecisionFixed 3
-            datetimeFormat = C.DatetimeFormat (Just '-') (Just ' ') (Just ':')
+    square
+    $ toStrict
+    $ TB.toLazyText
+    $ C.builder_DmyHMS timePrecision datetimeFormat (C.timeToDatetime t)
+  where
+    timePrecision = C.SubsecondPrecisionFixed 3
+    datetimeFormat = C.DatetimeFormat (Just '-') (Just ' ') (Just ':')
 
 showThreadId :: ThreadId -> Text
 showThreadId = square . T.pack . show
--- a/co-log/src/Colog/Message.hs
+++ b/co-log/src/Colog/Message.hs
@@ -321,7 +321,7 @@ Practically, it formats a message as 'fmtRichMessageDefault' without the severit
 -}
 fmtSimpleRichMessageDefault :: MonadIO m => RichMsg m SimpleMsg -> m Text
 fmtSimpleRichMessageDefault msg = fmtRichMessageCustomDefault msg formatRichMessage
-   where
+  where
     formatRichMessage :: Maybe ThreadId -> Maybe C.Time -> SimpleMsg -> Text
     formatRichMessage (maybe "" showThreadId -> thread) (maybe "" showTime -> time) SimpleMsg{..} =
         time
--- a/co-log/src/Colog/Message.hs
+++ b/co-log/src/Colog/Message.hs
@@ -271,7 +271,7 @@ defaultFieldMap = fromList
 
 -- | Contains additional data to 'Message' to display more verbose information.
 data RichMsg (m :: Type -> Type) (msg :: Type) = RichMessage
-    { richMsgMsg :: {-# UNPACK #-} !msg
+    { richMsgMsg :: !msg
     , richMsgMap :: {-# UNPACK #-} !(FieldMap m)
     } deriving (Functor)
 
--- a/co-log/src/Colog/Message.hs
+++ b/co-log/src/Colog/Message.hs
@@ -321,7 +321,7 @@ Practically, it formats a message as 'fmtRichMessageDefault' without the severit
 -}
 fmtSimpleRichMessageDefault :: MonadIO m => RichMessage m -> m Text
 fmtSimpleRichMessageDefault msg = fmtRichMessageCustom msg formatRichMessage
-   where
+  where
     formatRichMessage :: Maybe ThreadId -> Maybe C.Time -> Message -> Text
     formatRichMessage (maybe "" showThreadId -> thread) (maybe "" showTime -> time) Msg{..} =
         time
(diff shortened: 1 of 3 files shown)


co-log/src/Colog/Message.hs
```diff
@@ -262,10 +270,12 @@ defaultFieldMap = fromList
...
-data RichMessage (m :: Type -> Type) = RichMessage
-    { richMessageMsg :: {-# UNPACK #-} !Message
-    , richMessageMap :: {-# UNPACK #-} !(FieldMap m)
-    }
+data RichMsg (m :: Type -> Type) (msg :: Type) = RichMessage
+    { richMsgMsg :: !msg
+    , richMsgMap :: {-# UNPACK #-} !(FieldMap m)
+    } deriving (Functor)
+
+type RichMessage m = RichMsg m Message
```
chshersh:
```suggestion
-- | Specialised version of 'RichMsg' that stores severity, callstack and text message.
type RichMessage m = RichMsg m Message
```

co-log/src/Colog/Message.hs
```diff
@@ -283,10 +293,7 @@ __Examples:__
 See 'fmtMessage' if you don't need both time and thread id.
 -}
 fmtRichMessageDefault :: MonadIO m => RichMessage m -> m Text
-fmtRichMessageDefault RichMessage{..} = do
-    maybeThreadId  <- extractField $ TM.lookup @"threadId"  richMessageMap
-    maybePosixTime <- extractField $ TM.lookup @"posixTime" richMessageMap
-    pure $ formatRichMessage maybeThreadId maybePosixTime richMessageMsg
+fmtRichMessageDefault msg = fmtRichMessageCustom msg formatRichMessage
```
chshersh: Thanks for trying to cleanup code duplication! I've created another issue with the discussion about possible improvements to code modularity.

* https://github.com/kowainik/co-log/issues/129

co-log/src/Colog/Message.hs
```diff
@@ -296,18 +303,49 @@ fmtRichMessageDefault RichMessage{..} = do
...
+    maybePosixTime <- extractField $ TM.lookup @"posixTime" richMsgMap
+    pure $ formatter maybeThreadId maybePosixTime richMsgMsg
+
+showTime :: C.Time -> Text
+showTime t =
+            square
+          $ toStrict
+          $ TB.toLazyText
+          $ C.builder_DmyHMS timePrecision datetimeFormat (C.timeToDatetime t)
+        where
```
chshersh: `where` should have 2-space indentation and other functions should have 4-space indentation according to our style guide.

co-log/src/Colog/Message.hs
```diff
@@ -296,18 +303,49 @@ fmtRichMessageDefault RichMessage{..} = do
...
+fmtSimpleRichMessageDefault msg = fmtRichMessageCustom msg formatRichMessage
+   where
+    formatRichMessage :: Maybe ThreadId -> Maybe C.Time -> Message -> Text
+    formatRichMessage (maybe "" showThreadId -> thread) (maybe "" showTime -> time) Msg{..} =
+        time
+     <> showSourceLoc msgStack
+     <> thread
+     <> msgText
+
+fmtRichMessageCustom :: MonadIO m => RichMessage m -> (Maybe ThreadId -> Maybe C.Time -> Message -> Text) -> m Text
```
chshersh: This function is not modular enough. The problem is that with both `fmtRichMessageDefault` and `fmtSimpleRichMessageDefault` you give `RichMessage m` as an argument, where `fmtSimpleRichMessageDefault` should have the following type:

```haskell
fmtSimpleRichMessageDefault :: MonadIO m => RichMsg m SimpleMsg -> m Text
```

And then you can generalize type of this function to the following:

```haskell
fmtRichMessageCustomDefault
    :: MonadIO m 
    => RichMessage m msg 
    -> (Maybe ThreadId -> Maybe C.Time -> msg -> Text) 
    -> m Text
```

co-log/src/Colog/Message.hs
sphaso: Would you like me to work on #129 in this same PR or should we shoot for it later?

co-log/src/Colog/Message.hs
sphaso: I've implemented these changes. Trying to have `tutorials/Main.hs` to compile I've noticed that the various log actions (`logInfo` etc.) all accept a `Message` or `Msg`, but `SimpleMsg` is neither... should we try to make those functions more generic as well or am I chasing down a rabbit hole?

co-log/src/Colog/Message.hs
chshersh: @sphaso Much later. It's not even clearer to implement this 🤔 But if you would like to work with the advanced Haskell topics, let me know

co-log/src/Colog/Message.hs
```diff
@@ -296,18 +304,49 @@ fmtRichMessageDefault RichMessage{..} = do
...
+@
+[03 05 2019 05:23:19.058] [Main.example#34] [ThreadId 11] app: First message...
+[03 05 2019 05:23:19.059] [Main.example#35] [ThreadId 11] app: Second message...
+@
+
+Practically, it formats a message as 'fmtRichMessageDefault' without the severity information.
+-}
+fmtSimpleRichMessageDefault :: MonadIO m => RichMessage m -> m Text
+fmtSimpleRichMessageDefault msg = fmtRichMessageCustom msg formatRichMessage
+   where
```
chshersh: There should be 2 spaces before `where`
```suggestion
  where
```

co-log/src/Colog/Message.hs
chshersh: Well, `SimpleMsg` doesn't have `Severity` that's why it makes sense why you cannot use functions like `logInfo`. However, we have function `logMsg` which works with a polymorphic message:

https://github.com/kowainik/co-log/blob/a91ad4bdcc3a805e567ae124f5c4b729d95932df/co-log/src/Colog/Monad.hs#L56

I think it's better to create a new example with the `WithLog env Text m` constraint, so you can run it with your new functions.

sphaso: > Well, SimpleMsg doesn't have Severity

*facepalming hard*. Sorry, I must have been asleep.

I'm trying with creating a new example and having a 
`withLog (cmap fmtSimpleRichMessageDefault) simpleExample`
inside the `app` function of the tutorial.    
Getting a bit lost with the type tetris, but I'll get there :muscle:

chshersh: Keep up the good work! 👍

sphaso: @chshersh thanks! I've pushed another diff:    
- `RichMessage` was both a type and a constructor and GHC was getting quite confused, so I renamed the constructor `RichMessagex` just to make it compile. Do you have any idea on how to solve this? or do you have a better name?     
- as you can see, the approach I took in `play-colog` is a bit different from what was there before, I tried using the `withLog` function for days without getting much done. I don't like my current implementation (having to explicitly create a `RichMsg m SimpleMsg` is pretty awkward) but couldn't find a better way. Really hoping for you enlightenment :yin_yang:

co-log/tutorials/Main.hs
```diff
@@ -34,6 +36,13 @@ example = do
     log D "First message..."
     logInfo "Second message..."
 
+simpleExample :: MonadIO m => RichMsg m SimpleMsg
+simpleExample = RichMessagex (SimpleMsg callStack "Simple text message") defaultFieldMap
+
+simpleApp :: (MonadIO m, WithLog env Text m) => m ()
+simpleApp =
+    logMsg =<< fmtSimpleRichMessageDefault simpleExample
```
chshersh: I think that the example needs to be changed. Let me try to elaborate on how it should look like. 

The main idea behind `SimpleMsg` is that users of logging can provide only textual message and they will automatically see nice things like timestamp. thread id, line code. So instead of old:

```haskell
putStrLn "Some message..."
```

They can use:

```haskell
logMsg "Some message..."
```

And get a much nicer output. This should be possible in theory, because `CallStack` is acquired automatically from the context, and all IO actions are stored in the `FieldMap`. So let's try to achieve that.

First, the example we want to run should log only text message. It should be something like that:

```haskell
simpleMsgExample :: WithLog env SimpleMsg m => m ()
simpleMsgExample = do
    logText "First simple message"
    logText "Second simple message"
```

To run this example, we need to construct `LogAction IO Text` (as you did in `runSimpleApp`). But this action should be constructed differently, so it gets everything automatically. Basically, you need to write a function:

```
toSimpleMsg :: LogAction Text Simpl -> LogAction IO SimpleMsg
```

Which takes constructed `LogAction IO Text` (in our case `textAction` variable) and redirects output of the formatted `SimpleMsg` to `LogAction IO Text`. You also need to implement `logText` function manually which will construct `SimpleMsg` from `Text` by adding `CallStack` to it automatically.

sphaso: @chshersh Thank you a lot for your mentoring and patience. I'm working on it. There's one small piece of the puzzle I'm still missing. The formatter takes a `RichMsg m SimpleMsg` but here with `logText` we're building just a `SimpleMsg`. Is the formatter signature wrong? Should `logText` return a `RichMsg m SimpleMsg` instead of a simple `SimpleMsg`?

chshersh: @sphaso It's not a problem that formatter takes `RichMsg`. The idea here is that conversion to `RichMsg` happens implicitly and transparent for the user. You have:

```haskell
textAction :: LogAction IO Text
```

Then you are going to use [`cmapM`](https://hackage.haskell.org/package/co-log-core-0.2.0.0/docs/Colog-Core-Action.html#v:cmapM) function to create:

```haskell
richAction :: LogAction IO (RichMsg IO SimpleMsg)
richAction = cmapM fmtSimpleRichMessageDefault textAction
```

And then you can use `cmap` and `upgradeMessageAction` to create final action:

```haskell
simpleAction :: LogAction IO SimpleMsg
simpleAction = upgradeMessageAction defaultFieldMap richAction
```

> **NOTE:** For the final part you also need to generalise the type of the `upgradeRichMessage` function

And that's it. So the value flow is the following here:

1. You pass `Text` to `logText` function.
2. `logText` creates `SimpleMsg` from the `Text` and `CallStack` and outputs `SimpleMsg` to the underlying `LogAction`.
3. Than `RichMsg` is automatically created from `SimpleMsg` by attaching `defaultFieldMap` to it.
4. Then you format `RichMsg` as `Text`.
5. And then you output `Text` to the given `textAction` finally.

sphaso: Thank you so much! I think I've finally made it. I've pushed a new commit where I follow the guideline you provided. Let me know if you'd like anything to be changed.

PR #151, opened 2019-09-27
Bump version bounds

spacekitteh:


co-log-core/co-log-core.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:       2.4
 name:                co-log-core
-version:             0.2.0.0
+version:             0.2.0.1
```
chshersh: No need to bump up version here. The package can be revised on Hackage if only dependencies bump ups are required without code changes.

chshersh:
@Avi-D-coder No need to squash commits inside a branch :slightly_smiling_face: According to our workflow, we press `Squash and merge` button when merging PRs, so all commits are going to be automatically squashed when merging to `master`. Squashing inside a branch makes sense if you want to rebase on `master` — it will be easier to do with squashed commits.

co-log-core/co-log-core.cabal
spacekitteh: Alright, removed that change :)

chshersh:
Okay, now build fails because not all dependencies are supporting `ghc-8.8.1` at the moment. So we will wait when the dependencies are upgraded.

chshersh:
This PR solves the problem only partially, and this update should be performed more carefully and also after adding Windows CI support because we had issues with that before. We will take care of this by ourselves, so I'm closing this PR.

Issue #148, opened 2019-09-18
Support GHC-8.8.1


chshersh:
TODO:
- [ ] Update Travis config
- [ ] Add `-Wmissing-deriving-strategies` to `.cabal` file ghc options

PR #152, opened 2019-10-01
Support GHC 8.8.1

SanchayanMaity:
Implements #148 

Signed-off-by: Sanchayan Maity <maitysanchayan@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ language: haskell
 git:
   depth: 5
 
-cabal: "head"
+cabal: "3.0"
 
 cache:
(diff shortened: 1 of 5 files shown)


chshersh:
@SanchayanMaity Thanks for your PR! I see that it partially overlaps with PR #151. And also blocked by some upstream dependencies.

chshersh:
This PR solves the problem only partially, and this update should be performed more carefully and also after adding Windows CI support because we had issues with that before. We will take care of this by ourselves, so I'm closing this PR.

PR #154, opened 2019-10-07
[#138] Add higher-order transformation function

### LINKED ISSUE #138: Add higher-order transformation function
The following function might be useful

```haskell
hoistLogAction :: (forall x . m x -> n x) -> LogAction m a -> LogAction n a
```

jiribenes:
Resolves #138.

Haven't figured out a nice example/test for it. :/
Also, I had to add the `Rank2Types` language extension.

### DIFF
--- a/co-log-core/src/Colog/Core/Action.hs
+++ b/co-log-core/src/Colog/Core/Action.hs
@@ -1,4 +1,5 @@
-{-# LANGUAGE CPP #-}
+{-# LANGUAGE CPP        #-}
+{-# LANGUAGE Rank2Types #-}
 
 {- |
 Copyright:  (c) 2018-2019 Kowainik
@@ -48,6 +49,9 @@ module Colog.Core.Action
        , (<<=)
        , duplicate
        , multiplicate
+
+         -- * Higher-order combinators
+       , hoistLogAction
        ) where
 
 import Control.Monad (when, (>=>))
@@ -575,3 +579,10 @@ multiplicate (LogAction l) = LogAction $ \msgs -> l (fold msgs)
 {-# INLINE multiplicate #-}
 {-# SPECIALIZE multiplicate :: Monoid msg => LogAction m msg -> LogAction m [msg] #-}
 {-# SPECIALIZE multiplicate :: Monoid msg => LogAction m msg -> LogAction m (NonEmpty msg) #-}
+
+{- | Allows changing the internal monadic action -}
+hoistLogAction
+    :: (forall x. m x -> n x)
+    -> LogAction m a
+    -> LogAction n a
+hoistLogAction f (LogAction l) = LogAction (f . l)
--- a/co-log-core/src/Colog/Core/Action.hs
+++ b/co-log-core/src/Colog/Core/Action.hs
@@ -580,9 +580,27 @@ multiplicate (LogAction l) = LogAction $ \msgs -> l (fold msgs)
 {-# SPECIALIZE multiplicate :: Monoid msg => LogAction m msg -> LogAction m [msg] #-}
 {-# SPECIALIZE multiplicate :: Monoid msg => LogAction m msg -> LogAction m (NonEmpty msg) #-}
 
-{- | Allows changing the internal monadic action -}
+{- | Allows changing the internal monadic action.
+
+Let's say we have a pure logger action using 'PureLogger'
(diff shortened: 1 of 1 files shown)


co-log-core/src/Colog/Core/Action.hs
```diff
@@ -1,4 +1,5 @@
-{-# LANGUAGE CPP #-}
+{-# LANGUAGE CPP        #-}
+{-# LANGUAGE Rank2Types #-}
```
chshersh: It's okay to add `Rank2Types` :+1:

co-log-core/src/Colog/Core/Action.hs
```diff
@@ -575,3 +579,10 @@ multiplicate (LogAction l) = LogAction $ \msgs -> l (fold msgs)
 {-# INLINE multiplicate #-}
 {-# SPECIALIZE multiplicate :: Monoid msg => LogAction m msg -> LogAction m [msg] #-}
 {-# SPECIALIZE multiplicate :: Monoid msg => LogAction m msg -> LogAction m (NonEmpty msg) #-}
+
+{- | Allows changing the internal monadic action -}
+hoistLogAction
+    :: (forall x. m x -> n x)
+    -> LogAction m a
+    -> LogAction n a
+hoistLogAction f (LogAction l) = LogAction (f . l)
```
chshersh: Let's add an `{-# INLINE #-}` pragma to this function

co-log-core/src/Colog/Core/Action.hs
```diff
@@ -575,3 +579,10 @@ multiplicate (LogAction l) = LogAction $ \msgs -> l (fold msgs)
 {-# INLINE multiplicate #-}
 {-# SPECIALIZE multiplicate :: Monoid msg => LogAction m msg -> LogAction m [msg] #-}
 {-# SPECIALIZE multiplicate :: Monoid msg => LogAction m msg -> LogAction m (NonEmpty msg) #-}
+
+{- | Allows changing the internal monadic action -}
+hoistLogAction
```
chshersh: The example of this function is actually very simple. Let's say that you have pure logger action and you want to log all messages inside `IO`. So you have the following function:

```haskell
perfromPureLogsInIO :: PureLogger a -> IO a
```

You can then convert action that does logging in a pure monad to action that performs logging in the `IO` monad.

jiribenes: Thanks for clarification, I'll add this into the docs! :)

co-log-core/src/Colog/Core/Action.hs
jiribenes: Added in b75fc37.

jiribenes:
Thanks for the quick review! :)
I've tried addressing your suggestions and added an example!

co-log-core/src/Colog/Core/Action.hs
```diff
@@ -575,3 +579,28 @@ multiplicate (LogAction l) = LogAction $ \msgs -> l (fold msgs)
...
+
+@
+performPureLogsInIO :: PureLogger a -> IO a
+@
+
+then we can convert a logger action that uses a pure monad
+to a one that performs the logging in the 'IO' monad using:
+
+@
+hoistLogAction performPureLogsInIO :: LogAction (PureLogger a) a -> LogAction IO a
```
chshersh: Minor documentation typo :slightly_smiling_face: 
```suggestion
hoistLogAction performPureLogsInIO :: LogAction PureLogger a -> LogAction IO a
```

jiribenes: Sorry, just to better understand your point:

I thought that `PureLogger` in the `Colog.Pure` module always has a message argument to go along with it, for example here: https://github.com/kowainik/co-log/blob/e848fed2939a29c4293f6fcf5de6f547882e4d4a/co-log/test/Property.hs#L27

So that's why I'd assume that the type of a pure logger action should be `LogAction (PureLogger msg) msg`.

Could you please explain why is my understanding incorrect? :)

chshersh: @jiribenes Your understanding is 100% correct :ok_hand: I just had too much PRs to review and missed some parts of the implementation. There's no type-checker for the documentation, so it won't harm to be extra careful.

jiribenes:
Thanks for the quick review! :heart:

PR #161, opened 2019-12-27
[#124] Implement concurrent playground for logging

### LINKED ISSUE #124: Implement playground for concurrent testing of logging
Just to check whether the output is interleaved or not.

chshersh:
Resolves #124

I tested it locally, I don't see any interleaved output. I see the following correct output:

```
Logging from thread: 4
Logging from thread: 5
Logging from thread: 3
Logging from thread: 6
Logging from thread: 1
Logging from thread: 2
Logging from thread: 10
Logging from thread: 8
Logging from thread: 4
Logging from thread: 6
Logging from thread: 7
Logging from thread: 2
Logging from thread: 10
Logging from thread: 8
Logging from thread: 4
Logging from thread: 9
Logging from thread: 6
Logging from thread: 2
Logging from thread: 10
```

So if somebody has problems with concurrent logging, we at least have a place to experiment with such problems...


chshersh:
@vrom911 good idea 👍 I will update the PR.

PR #167, opened 2019-12-30
[#122] Add separate combinator

### LINKED ISSUE #122: Add a function to log batch of messages independently
There is comonadic operator `multiplicate` that combines any number of messages into a single message and then logs singe message. It has the following implementation:

https://github.com/kowainik/co-log/blob/746016089a44e06dadf23818044285bd3ab22135/co-log-core/src/Colog/Core/Action.hs#L569-L574

It would be nice to have a function that can log any `Foldable` of messages by logging each message independently. This function has relation with the `Divisible` typeclass. 

```haskell
??? :: (Traversable f, Applicative m) => LogAction m msg -> LogAction m (f msg)
??? (LogAction action) = LogAction (traverse_ action)
```

**Unresolved question:** Name for the function.

On the other hand, it's just `contramap traverse_`....

vrom911:
Resolves #122

### DIFF
--- a/co-log-core/src/Colog/Core/Action.hs
+++ b/co-log-core/src/Colog/Core/Action.hs
@@ -624,15 +624,15 @@ multiplicate (LogAction l) = LogAction $ \msgs -> l (fold msgs)
 of them separately.
 
 >>> :{
-let logger :: LogAction IO [Int]
+let logger :: LogAction IO Int
     logger = logPrint
-in separate logger <& replicate 5 [1..3]
+in separate logger <& [1..5]
 :}
-[1,2,3]
-[1,2,3]
-[1,2,3]
-[1,2,3]
-[1,2,3]
+1
+2
+3
+4
+5
 
 @since 0.2.1.0
 -}
--- a/co-log-core/src/Colog/Core/Action.hs
+++ b/co-log-core/src/Colog/Core/Action.hs
@@ -53,6 +53,7 @@ module Colog.Core.Action
        , (<<=)
        , duplicate
        , multiplicate
+       , separate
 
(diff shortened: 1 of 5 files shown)


co-log-core/src/Colog/Core/Action.hs
```diff
@@ -619,6 +620,34 @@ multiplicate (LogAction l) = LogAction $ \msgs -> l (fold msgs)
 {-# SPECIALIZE multiplicate :: Monoid msg => LogAction m msg -> LogAction m [msg] #-}
 {-# SPECIALIZE multiplicate :: Monoid msg => LogAction m msg -> LogAction m (NonEmpty msg) #-}
 
+{- | Like 'multiplicate' but instead of logging a batch of messages it logs each
+of them separately.
+
+>>> :{
+let logger :: LogAction IO [Int]
+    logger = logPrint
+in separate logger <& replicate 5 [1..3]
```
chshersh: Let's have a bit different example for easier understanding. Let's have `logger :: LogAction IO Int` and `separate logger <& [1..5]`. I think when types are `Int` and `[Int]` vs. `[Int]` and `[[Int]]`, it's a bit easier to understand what's going on in the example.

co-log-core/test/Doctests.hs
```diff
@@ -7,7 +7,7 @@ import Test.DocTest (doctest)
 
 main :: IO ()
 main = do
-    sourceFiles <- glob "co-log-core/src/**/*.hs"
+    sourceFiles <- glob "src/**/*.hs"
```
chshersh: Thanks for fixing `doctest`!

PR #169, opened 2020-01-02
[#139] Add type-errors for `LogAction` instances

### LINKED ISSUE #139: Add type-errors for `LogAction` instances
I thought of an idea that custom type errors can improve the quality of `co-log-core` even further. For example, we can add instances with type errors for `Functor`, `Applicative` and `Monad` classes just in case somebody uses wrong abstractions.

vrom911:
Resolves #139

instances for `Applicative` and `Monad` cannot be created as they imply `Functor` instance which immediately gives introduced here custom type error.

### DIFF
--- a/co-log-core/CHANGELOG.md
+++ b/co-log-core/CHANGELOG.md
@@ -8,7 +8,7 @@ The change log is available [on GitHub][2].
 * [#122](https://github.com/kowainik/co-log/issues/122):
   Add the `separate` combinator.
 * [#139](https://github.com/kowainik/co-log/issues/139):
-  Add (unrepresentable) 'Functor' instance for 'LogAction' with the custom type-error.
+  Add (unrepresentable) `Functor` instance for `LogAction` with the custom type-error.
 
 ## 0.2.0.0 — May 5, 2019
 
--- a/co-log-core/CHANGELOG.md
+++ b/co-log-core/CHANGELOG.md
@@ -7,6 +7,8 @@ The change log is available [on GitHub][2].
 
 * [#122](https://github.com/kowainik/co-log/issues/122):
   Add the `separate` combinator.
+* [#139](https://github.com/kowainik/co-log/issues/139):
+  Add (unrepresentable) 'Functor' instance for 'LogAction' with the custom type-error.
 
 ## 0.2.0.0 — May 5, 2019
 
--- a/co-log-core/src/Colog/Core/Action.hs
+++ b/co-log-core/src/Colog/Core/Action.hs
@@ -184,7 +184,7 @@ In case it is used by mistake, the user  will see the following:
 >>> fmap show logStringStdout
 ...
 ... 'LogAction' cannot have a 'Functor' instance by design.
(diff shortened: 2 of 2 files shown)


co-log-core/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ The change log is available [on GitHub][2].
 
 * [#122](https://github.com/kowainik/co-log/issues/122):
   Add the `separate` combinator.
+* [#139](https://github.com/kowainik/co-log/issues/139):
+  Add (unrepresentable) 'Functor' instance for 'LogAction' with the custom type-error.
```
chshersh: Minor formatting suggestion :slightly_smiling_face: 
```suggestion
  Add (unrepresentable) `Functor` instance for `LogAction` with the custom type-error.
```

co-log-core/src/Colog/Core/Action.hs
```diff
@@ -144,13 +150,68 @@ instance Applicative m => Monoid (LogAction m a) where
 
 #if MIN_VERSION_base(4,12,0)
 instance Contravariant.Contravariant (LogAction m) where
+    contramap :: (a -> b) -> LogAction m b -> LogAction m a
     contramap = cmap
     {-# INLINE contramap #-}
 
+    (>$) :: b -> LogAction m b -> LogAction m a
```
chshersh: Thanks for adding type signatures! This is really helpful here :+1:

co-log-core/src/Colog/Core/Action.hs
```diff
@@ -144,13 +150,68 @@ instance Applicative m => Monoid (LogAction m a) where
...
+        ':$$: 'Text "  * contramap :: (a -> b) -> LogAction m b -> LogAction m a"
+        ':$$: 'Text "  * (>$) :: b -> LogAction m b -> LogAction m a"
+#endif
+        )
+
+{- | ⚠️__CAUTION__⚠️ This instance is for custom error display only.
+
+'LogAction' is not supposed to have 'Functor' instance by design.
+
+In case it is used by mistake, the user  will see the following:
```
chshersh: :heart: :sweat_drops:

co-log-core/src/Colog/Core/Action.hs
```diff
@@ -144,13 +150,68 @@ instance Applicative m => Monoid (LogAction m a) where
...
     (>$) = (Colog.Core.Action.>$)
     {-# INLINE (>$) #-}
 #endif
 
+-- | For tracking usage of unrepresentable class instances of 'LogAction'.
+type family UnrepresentableClass :: Constraint
+  where
+    UnrepresentableClass = TypeError
+        ( 'Text "'LogAction' cannot have a 'Functor' instance by design."
+        ':$$: 'Text "However, the instance is used in here."
```
chshersh: Just a minor rewording, but I'm not sure about the correct words. This is already great :+1: 
```suggestion
        ':$$: 'Text "However, you've attempted to use this instance."
```

PR #174, opened 2020-01-19
Update Main.hs

mb720:
I needed to add ScopedTypeVariables to get this to compile:

    let runApp :: LogAction IO Message -> IO ()

### DIFF
--- a/co-log/tutorials/Main.hs
+++ b/co-log/tutorials/Main.hs
@@ -8,6 +8,7 @@
 {-# LANGUAGE MultiParamTypeClasses      #-}
 {-# LANGUAGE OverloadedStrings          #-}
 {-# LANGUAGE PatternSynonyms            #-}
+{-# LANGUAGE ScopedTypeVariables        #-}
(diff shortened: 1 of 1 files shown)


chshersh:
@mb720 Can you share how exactly do you compile? The `ScopedTypeVariables` extension is enabled in the `default-extensions` section of `co-log.cabal`, so you don't need to specify it in a module. I'm sure that the current code compiles as we check it on CI. That's why I'm asking for clarification :slightly_smiling_face:

mb720:
Ah, I see.

I didn't use `co-log.cabal`, instead I copied the code into my own project.

Have you considered moving the default language extensions from `co-log.cabal` int the `.hs` files? This way the files work on their own and can be copy-pasted, like I did.

chshersh:
@mb720 The `Main.hs` module is a part of a project. If it was a single self-contained script, then it would have been written in a completely different way as now copying the file will require a lot more than just adding an extension. But we don't have a goal to convert it into a runnable singleton file. I'm closing this PR since this doesn't satisfy goals.

PR #205, opened 2020-08-25
[#197] Remove strict annotation of msgText

### LINKED ISSUE #197: `filterBySeverity` may bring unexpected extra computation on `Msg` due to strictness?
Sometimes we want to filter some log messages by `filterBySeverity`. What we expect is that if a messages is filtered, the action will not be executed at all. However, the pre-defined message types `Msg`, `SimpleMsg` and `Message` have a [strict `!Text` field](https://github.com/kowainik/co-log/blob/17d91e47cf3850a9e9f879f2bd9c01e661f13e28/co-log/src/Colog/Message.hs#L109), which can bring unexpected computation on message we want to log.

For example:

```haskell
level :: MonadIO m
      => Colog.LogAction m Colog.Message
level = Colog.filterBySeverity Colog.I sev Colog.richMessageAction
  where sev Colog.Msg{..} = msgSeverity

app :: App ()
app = do
  let veryLongMsg = Text.replicate 100000000 "aaaaa"
  Colog.logDebug veryLongMsg
  Colog.logInfo  "bbbbb"
```

In this example, I do not want `logDebug` to be executed. However due to the strictness, `veryLongMsg` will be computed instead of being lazy. This slightly goes against the semantics of `filterBySeverity` and can bring performance problems sometimes.

So I want to know if the strict mark is for special usage and if the behavior above is a bug or just an improper use of the library.

Commelina:
Resolves #197

### DIFF
--- a/co-log/src/Colog/Message.hs
+++ b/co-log/src/Colog/Message.hs
@@ -106,7 +106,7 @@ locations where each particular function was called.
 data Msg sev = Msg
     { msgSeverity :: !sev
     , msgStack    :: !CallStack
-    , msgText     :: !Text
+    , msgText     ::  Text
     }
 
 {- | Message data type without 'Severity'. Use 'logText' to log


chshersh (approved):

chshersh:
Thanks!

PR #216, opened 2021-02-25
Show local timezone instead of UTC?

alaendle:
First I need to apologize because this is more a question than a pull request; but it's so much easier to communicate an idea with some accompanying code 😀 

So, should co-log always use UTC as time zone output? 
Personally I'm unsure what is the "correct" behaviour, so I started this experiment to use the local time zone as an default instead.

Please note that this adds the `time` package as a dependency to `co-log` (but it was a transitive dependency before, so it shouldn't add to the footprint of the library).

So just tell me what do you think?


alaendle:
Superseeded by https://github.com/co-log/co-log/pull/226.

PR #221, opened 2021-06-11
replace symlinks with files

teto:
when using haskell.nix (but I suspect the same issue may arise with more cabal tooling), it downloads only the subdir co-log-polysemy since that's what I use and thus the symlinks are left dangling: https://github.com/input-output-hk/haskell.nix/issues/1134

I replaced the symlinks with a copy of the file. I dont know if that's the proper fix though.

### DIFF
--- a/co-log-polysemy/LICENSE
+++ b/co-log-polysemy/LICENSE
@@ -1,1 +1,373 @@
-../LICENSE
+Mozilla Public License Version 2.0
+==================================
+
+1. Definitions
+--------------
+
+1.1. "Contributor"
+    means each individual or legal entity that creates, contributes to
+    the creation of, or owns Covered Software.
+
+1.2. "Contributor Version"
+    means the combination of the Contributions of others (if any) used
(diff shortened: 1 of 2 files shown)


sorki:
I've hit the same issue and created a commit that copies all of them  :smiley_cat: https://github.com/sorki/co-log/commit/5abf5ab0b6f665c0b282e519c8b85da94d88b263

chshersh:
@teto Thanks for your contribution. However, we don't have the capacity to maintain the duplicates of README files for every package each time we change the documentation. That's why we're using a symlink approach. Duplicating LICENSE is fine because it doesn't change but it doesn't solve the problem entirely so I don't see much benefit in deduplicating only LICENSE.

If you're using Nix, you can see how the symlink problem was solved in Summoner previously:

* https://github.com/kowainik/summoner/pull/335/files#diff-245563af9e34754739b12b294f3324c1d42c116ace7ce8735e21dfbc82d8e064R19-R26

PR #223, opened 2021-08-14
GHC 9 support

sorki:
Relaxes `base` bounds, bumps `typerep-map` and `polysemy`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-16.04
     strategy:
       matrix:
-        cabal: ["3.0"]
+        cabal: ["3.4"]
         ghc:
           - "8.2.2"
           - "8.4.4"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,7 @@ jobs:
           - "8.6.5"
           - "8.8.3"
           - "8.10.1"
+          - "9.0.1"
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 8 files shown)


vrom911 (commented):
Hey @sorki ,

Thanks a lot for this update! Looks nice!
Could you please also add `9.0` into the GitHub Actions file, so we could have a check on CI that everything is still good?

sorki:
Sure, added! I think it's going to get cancelled due to some unrelated fail in older one(s) but lets see.

sorki:
All green now.

chshersh (approved):
Thanks! Looks neat 🙂

vrom911 (approved):
Thanks!

chshersh:
@alaendle FYI, [`co-log-polysemy-0.0.1.3`](https://hackage.haskell.org/package/co-log-polysemy-0.0.1.3) has been released to Hackage with GHC 9.0 support.

alaendle:
@chshersh Many thanks for the hint, unfortunately my project not only depends in `co-log-polysemy`, but only on `co-log` - so since their latest releases on hackage differ in the version dependency on `co-log-core` ((>=0.3.0.0 && <0.4) vs. (>=0.2.1.0 && <0.3)) I'm still using the git repository

```yaml
extra-deps:
- git: https://github.com/co-log/co-log.git
  commit: 72fbe394b437c698d574cd7604ad3f7f807383e0
  subdirs:
    - co-log
    - co-log-core
    - co-log-polysemy
```
for now. This is only for explanation and if someone experiences a similar problem and in absolute no way criticism! I really appreciate your work and the design of the `co-log` library in particular 👍 .

PR #228, opened 2022-02-01
Simple (!) integration of doctests for co-log.

alaendle:


chshersh (commented):
This looks okay. But usually, in all our projects we define Doctests in a separate test stanza to not mix doctest with other tests. Check the example in `relude`:

* https://github.com/kowainik/relude/blob/715afb764034d88b2db2b7d69e08573589fdd41d/relude.cabal#L253-L263

I would like to have the same setup here as well.
Feel free to implement in a separate PR if you feel like it would be easier to just create a new PR 👍🏻

alaendle:
Your wish is my command 😉

co-log/src/Colog/Message.hs
```diff
@@ -503,7 +503,6 @@ builderDmyHMSz (C.OffsetDatetime (C.Datetime date time) offset) =
 
 {- | Shows a thread id in the following format:
 
->>> showThreadId <$> Control.Concurrent.myThreadId
```
chshersh, line 506: You don't need to delete this line. The documentation still can be helpful 🙂 
Usual trick in such situations is to replace `>>>` with `>>` and optionally make it **bold** for prettier printing 💅🏻 

```haskell
__>>__ showThreadId <$> Control.Concurrent.myThreadId

chshersh (commented):
Looks pretty good! 🔝 
I have only one comment. But the rest looks good 🙂

co-log/src/Colog/Message.hs
alaendle: Makes sense; wasn't aware of this trick.

chshersh (approved):

PR #236, opened 2022-02-24
Just use an existing path to be able to start the play-colog example …

alaendle:
…executable without an exception.


chshersh (approved):
Thanks for the fix!

PR #240, opened 2022-03-29
Add MonadUnliftIO instance

newhoggy:


alaendle:
Could we please merge this? It makes really sense to support this (guess this issue always pops up in advanced scenarios where a monad transformer stack is build). Sure - this brings in a new dependency - but `unliftio-core` is super lightweight.

alaendle (approved):

alaendle:
@newhoggy Many thanks for your contribution!

PR #243, opened 2022-05-21
Fix logs printing in multiple threads

worm2fed:
Problem:
Function to print text with newline which is in use in library performs
it as two separate actions - first is to print message and second - to
print newline symbol.
This way there is issues with computations in multiple threads, so
newline symbol sometimes printed in wrong place.

Solution:
Redefine logger action to use version where newline symbol appended
before printing.

### DIFF
--- a/src/Colog/Actions.hs
+++ b/src/Colog/Actions.hs
@@ -27,7 +27,7 @@ module Colog.Actions
 
 import Control.Monad.IO.Class (MonadIO (..))
 import Data.Text.Encoding (encodeUtf8)
-import System.IO (Handle, IOMode (AppendMode), stderr, withFile)
+import System.IO (Handle, IOMode (AppendMode), stderr, stdout, withFile)
 
 import Colog.Core.Action (LogAction (..), cmapM, (>$<))
 import Colog.Core.IO (logFlush)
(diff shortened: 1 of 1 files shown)


vrom911 (approved):
Looks great! Thanks!

worm2fed:
@vrom911 pipeline failed, sorry - created this PR via browser and forgot to import `stdout`

Also, there can be same issue with other actions (like `ByteString`), but I didn't test it

worm2fed:
@vrom911 seems like you need merge it by yourself:

> Merging is blocked
The base branch restricts merging to authorized users. [Learn more about protected branches.](https://docs.github.com/articles/about-protected-branches/)

alaendle:
Maybe something that also should also be reported upstream (https://github.com/haskell/text/issues)? I'm not sure about the promises that **text** makes about concurrency - and for sure this might also depend on the used BufferMode, however it is astonishing to have a difference between `hPutStr handle $ msg <> "\n"` and `hPutStrLn handle msg` - this just feels wrong (even if it might be o.k. from a library perspective that doesn't assume concurrent executions).

worm2fed:
https://github.com/haskell/text/issues/442

alaendle:
@vrom911 - Seems like you are the only person that is allowed to merge these changes. 
grafik
Since you've already approved the changes I think that would be fine.

vrom911:
Thanks a lot, merged! ❤️

Issue #134, opened 2019-06-07
Implement more `mtl`-style instances for LoggerT


jacobstanley:
It would be great if `LoggerT` supported some more of the more common `mtl`-style typeclasses, such as `MonadReader`, `MonadState`, `MonadThrow`, `MonadCatch`, `MonadMask`, `MFunctor`, `MonadResource`.

This isn't a huge show stopper because it's easy enough to implement your own `LoggerT` and/or integrate it in to an existing `ReaderT`, but it would be handy nonetheless.

jacobstanley:
I may be missing something about how you're supposed to use this library though!

pangwa:
> I may be missing something about how you're supposed to use this library though!

Actually I was just looking for a logging library which supports file &terminal logging and log file rotation. I didn't find any logging library supports natively except this one... But co-log's log rotation function doesn't work as I just tested.

chshersh:
@jacobstanley thanks for your suggestion! It would be indeed great to have many instances. However, this is tricky.

> MonadReader

`LoggerT` already implements `MonadReader` instance for `LogAction` and `MonadReader` typeclass has a functional dependency, that's why it's not possible to implement another `MonadReader` instance for it.

> MonadState, MonadWriter

Should be possible and very nice to have those! However, these instances cannot be derived automatically. Manual implementation is tricky, and I won't be confident in them without property-based tests on laws.

> MonadThrow, MonadCatch, MonadMask, MFunctor, MonadResource

I'm reluctant to add those constraints because they will increase dependency footprint of the `co-log` package and I don't want to go this way unless it's necessary.

> This isn't a huge show stopper because it's easy enough to implement your own LoggerT and/or integrate it in to an existing ReaderT, but it would be handy nonetheless.
> I may be missing something about how you're supposed to use this library though!

This is indeed how the library is supposed to be used: have your own environment and put `LogAction` inside it. This use case is covered in details in the corresponding tutorial:

* https://github.com/kowainik/co-log/blob/master/co-log/tutorials/2-custom/Custom.md

`LoggerT` is helpful only in the following scenarios:

1. When you want to replace `IO` with _FancyIO_ and use pretty logging instead of `putStrLn` and nothing else.
2. When you need pure logging.
3. `LoggerT` illustrates that `co-log` supports MTL-style architecture.

@pangwa 

[...]

PR #244, opened 2022-09-09
Derive `MonadFail` for `LoggerT`

NicolasT:
Not a complete fix for #134, but at least provide one more instance for `base` classes. Given `ReaderT` provides instances for many more classes (many of them in `base` or `mtl` which are dependencies already), more could be added.

See: https://github.com/co-log/co-log/issues/134


alaendle (approved):

alaendle:
Since I wasn't able to edit this PR and I wanted to do a forward integration before to check if everything still compiles I've created #260.

alaendle:
Mewrged in #260.

PR #253, opened 2023-04-03
docs: use relative path to benefit locally reading

xieyuschen:
I change the tutorial link to a relative path to benefit local reading so it will jump inside the viewer instead of opening a new web page.


vrom911 (approved):
Awesome! Thanks very much 👍🏼

PR #257, opened 2023-07-02
Update CI tested GHC versions & workaround for GHC < 9.4.5 (run-st, primitive-unlifted)

alaendle:


vrom911 (commented):
I aagree, Maybe we should add the reason in comments or docs somewhere explaining what is the blocker for current Windows test build? Otherwise, I agree 👍🏼

alaendle:
@vrom911 to be honest I couldn't really tell what goes wrong on Windows - locally I recognize that qualified imports didn't seem to work in the tests.
```
src\Colog\Message.hs:422: failure in expression `showTime $ C.Time 1577656800000000000'
expected: "[29 Dec 2019 22:00:00.000 +00:00] "
 but got:
          ^
          <interactive>:27:12: error:
              Not in scope: data constructor `C.Time'
              NB: no module named `C' is imported.
```
But also after using full names in the test code it didn't worked; also I believe this might be triggered by some change in `ghci`, but I haven't found a corresponding issue on the GHC issue tracker. So I just added a generell comment why `doctest` is disabled on windows. However I believe this shouldn't be a problem, since I think we are good enough if all tests pass on linux systems 😉

vrom911 (approved):

PR #259, opened 2023-08-17
docs: add a link to tutorial pages and aggregate all tutorial pages

xieyuschen:
- aggregated all the tutorial pages together and put a link to the tutorial readme.
- format the markdown using the `Prettier` plugin.
- fix: previously, the second link with the title *Using custom monad that stores LogAction inside its environment* is wrong and caused a 404. I apologize for such a careless mistake and fixed it in this PR.

Regards


vrom911 (approved):
Awesome!

Issue #129, opened 2019-05-22
[RFC] Combine formatting with FieldMap


chshersh:
Currently `FieldMap` contains only actions that extract additional fields. And formatting for them is done in a separate function. I think it can be more convenient to have everything in one place, when you specify, how to extract field and how to format it.

alaendle:
Really sounds beneficial - will create a draft to see how this might look like.

PR #271, opened 2023-11-03
First step towards combined formatting (#129).

alaendle:
PoC - DO NOT MERGE - Just for review - incomplete for now!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 dist
 dist-*
 cabal-dev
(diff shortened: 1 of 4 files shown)


xieyuschen (approved):
It's reasonable to combine the fmt with the concrete message to me as well. As I'm not sure the merits of `type family`, I'm trying to the real usages about it and thinking about whether it's good to replace it.

PR #272, opened 2024-01-08
docs: refine the loggert tutorials

xieyuschen:
# Changes
- add a `LoggerT` monad transformer usage in example
Previously, the tutorial uses the `WithLog env SimpleMsg m => m ()` function only for `usingLoggerT`. In my opinion, it's better for the tutorial to state the `WithLog env SimpleMsg m => m ()` and `LoggerT SimpleMsg m ()` is equivalent. 
Hence, I refined this tutorial.
- update the docs to run the tutorial as the cabal file requires a new flag

# Note
- the `LoggerT` example cannot print the location correctly. I guess it's normal as inside co-log does additional work for it. I haven't checked further so please correct me if i'm wrong and I will fix the example as well.

Regards


alaendle:
Hello @xieyuschen, thanks for your contribution and please excuse the late reply. To be honest I'm not really sure if this improves the documentation around `LoggerT` - on the one hand, yes, it really tries to build up examples around the transformer; on the other hand, no, it complicates things for newcomer and might be confusing. Unfortunately I still haven't found time to take a deeper look; but I also don't want to delay a reaction further - so please take this as some kind of interim feedback. Regarding the missing location - WithLog enforces HasCallStack - just add HasCallStack as an additional constraint and the location of the log statement should get visible.

alaendle (approved):

alaendle:
@xieyuschen - one again, thanks for your contribution - I've changed a few small things - if you still like it, I think we should merge it 👍 otherwise feel free to further improve this PR 😉

xieyuschen:
> Hello @xieyuschen, thanks for your contribution and please excuse the late reply. To be honest I'm not really sure if this improves the documentation around `LoggerT` - on the one hand, yes, it really tries to build up examples around the transformer; on the other hand, no, it complicates things for newcomer and might be confusing. Unfortunately I still haven't found time to take a deeper look; but I also don't want to delay a reaction further - so please take this as some kind of interim feedback. Regarding the missing location - WithLog enforces HasCallStack - just add HasCallStack as an additional constraint and the location of the log statement should get visible.

Hi @alaendle . sorry for late reply. The motivation of this tutorial is to show the equivalent of `LoggerT` and  the monad decorated by `WithLog`. When I tried to learn about the colog library, I got confused why it provides a `LoggerT` but rarely uses it in the example. Instead, we use the `WithLog` more frequently.

I'm not sure whether I miss some haskell knowledges that leads the confusion, or colog misses some documentation about them. I prefer the latter, hence I created a PR for appending this detail.

xieyuschen:
> @xieyuschen - one again, thanks for your contribution - I've changed a few small things - if you still like it, I think we should merge it 👍 otherwise feel free to further improve this PR 😉

sure, thanks for your kindly help. I will check it:)

alaendle:
I hope it was okay that I already merged - it's better to have a lot of small progress rather than being stuck 😃

PR #277, opened 2025-03-25
Move away from `chronos`.

### LINKED ISSUE #190: Consider moving away from `chronos`?
The library seems a bit unstable on different platforms, and some [fixes like this](https://github.com/andrewthad/chronos/pull/58) look a bit suspicious.

I propose to at least discuss options we have to improve this situation.

alaendle:
Would be a breaking change, but simplifies code and reduces dependencies (and therefore closes #190).

But we would get slower - around factor 3 - is this acceptable?
grafik
(also take into account that these are the times for 10k messages)

### DIFF
--- a/co-log.cabal
+++ b/co-log.cabal
@@ -113,7 +113,7 @@ library
                      , filepath >= 1.4.1 && < 1.6
                      , mtl >= 2.2.2 && < 2.4
                      , text >= 1.2.3 && < 2.2
-                     , chronos ^>= 1.1 && < 1.2
+                     , time >= 1.8 && < 1.15
                      , transformers >= 0.5 && < 0.7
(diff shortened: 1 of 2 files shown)


alaendle:
After some optimisation ... not as bad as before - but still more than double the time ...
grafik
But I guess this is more or less the maximum that could be achieved, since profiling shows that time formatting is responsible for around 76% of the needed time.

alaendle:
@F-Joachim Thoughts? 🤔 Trading less code and less dependencies for performance? At the moment I tend to merge this - as in the long-term simplicity always wins? (also performance might improve again, if we work on #129 - but not sure if this will happen).

F-Joachim:
Hi @alaendle - As already discussed, I would also merge it. I also believe that reducing the code and dependencies is a crucial argument.

